PR #29, opened 2023-08-10
Demo notebooks

ziw-liu:
TODO:

- [ ] Remove hard-coded file path.

### DIFF
--- a/examples/demo_dlmbl/python/excercise_1.py
+++ b/examples/demo_dlmbl/python/excercise_1.py
@@ -177,6 +177,8 @@ Increase the ``depth`` in ``draw_graph`` to zoom in.
 # %%
 model_config = {
     "architecture": "2D",
+    "in_channels": 1,
+    "out_channels": 2,
     "residual": True,
     "dropout": 0.1,
     "task": "reg",
--- a/examples/demo_dlmbl/python/excercise_1.py
+++ b/examples/demo_dlmbl/python/excercise_1.py
@@ -30,6 +30,7 @@ from viscy.light.data import HCSDataModule
 from viscy.light.engine import VSTrainer, VSUNet
 
 BATCH_SIZE = 32
+GPU_ID = 0
 
 # %% [markdown]
 """
@@ -193,7 +194,7 @@ model = VSUNet(
 )
 
 # visualize graph
-model_graph = draw_graph(model, model.example_input_array, depth=2)
+model_graph = draw_graph(model, model.example_input_array, depth=2, device="cpu")
 graph = model_graph.visual_graph
 graph
 
@@ -204,7 +205,7 @@ Here we use the ``fast_dev_run`` flag to run a sanity check first.
 """
 
 # %%
-trainer = VSTrainer(accelerator="gpu", fast_dev_run=True)
+trainer = VSTrainer(accelerator="gpu", device=GPU_ID, fast_dev_run=True)
 
 trainer.fit(model, datamodule=data_module)
 
--- a/examples/demo_dlmbl/python/excercise_1.py
+++ b/examples/demo_dlmbl/python/excercise_1.py
@@ -130,16 +130,33 @@ for i, batch in enumerate(train_dataloader):
 # %% tags=["solution"]
 train_dataloader = data_module.train_dataloader()
 
+
+fig, axs = plt.subplots(3, 8, figsize=(20, 6))
+
+# Draw 8 batches, each with 32 images. Show the first image in each batch.
+
 for i, batch in enumerate(train_dataloader):
-    print(f"Batch {i}:")
-    for k, v in batch.items():
-        if isinstance(v, torch.Tensor):
-            summary = (v.shape, v.dtype)
-        else:
-            summary = v
-        print(k, summary)
-    if i > 2:
+    # The batch is a dictionary consisting of three keys: 'index', 'source', 'target'.
+    if i >= 8:
         break
+    FOV = batch['index'][0][0]
+    input_tensor = batch['source'][0, 0, :, :].squeeze()
+    target_membrane_tensor = batch['target'][0, 0, :, :].squeeze()
+    target_nuclei_tensor = batch['target'][0, 1, :, :].squeeze()
+
+    axs[0, i].imshow(input_tensor, cmap='gray')
+    axs[1, i].imshow(target_nuclei_tensor, cmap='gray')
+    axs[2, i].imshow(target_membrane_tensor, cmap='gray')
+    axs[0, i].set_title(f'input@{FOV}')
+    axs[1, i].set_title('target-nuclei')
+    axs[2, i].set_title('target-membrane')
+    axs[0, i].axis('off')
+    axs[1, i].axis('off')
+    axs[2, i].axis('off')
+
+plt.tight_layout()
+plt.show()
+
 
 # %% [markdown]
 """
--- a/examples/demo_dlmbl/python/excercise_1.py
+++ b/examples/demo_dlmbl/python/excercise_1.py
@@ -55,9 +55,7 @@ Run <code>open_ome_zarr?</code> in a cell to see the docstring.
 
 # %%
 # set dataset path here
-data_path = (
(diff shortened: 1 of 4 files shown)


examples/demo_dlmbl/python/excercise_1.py
```diff
@@ -56,16 +59,25 @@
...
 as some cells are not expressing the fluorophore.
 """
 
 # %%
-image = dataset["0/0/0/0"].numpy()
+
+row = "0"
+col = "0"
+field = "0"
+resolution = "0" # 0 is the highest resolution, 1 is 2x2 binned, 2 is 4x4 binned, etc.
```
ziw-liu, line 79: In this dataset we will only be providing the full resolution. Will this be misleading?

mattersoflight: We should clarify the data format. The extra level of heirarchy in HCS format stumps a few people. Please add more resolutions using `ndpyramid` (https://pypi.org/project/ndpyramid/) in this dataset. We won't load these, but having them in dataset will clarify the format further.

examples/demo_dlmbl/python/excercise_1.py
```diff
@@ -0,0 +1,257 @@
...
+"""
+
+# %%
+train_dataloader = data_module.train_dataloader()
+
+for i, batch in enumerate(train_dataloader):
+    ...
+    # check some batches and break
+    break
+
```
mattersoflight, line 133: What are the tensors located at batch['index'][1] and batch['index'][2]?

examples/demo_dlmbl/python/excercise_1.py
ziw-liu: `ndpyramid` seems abandoned and does not support OME-Zarr. Since this is a 2D dataset, I will just use the [iohub method](https://czbiohub-sf.github.io/iohub/main/api/ngff.html#iohub.ngff.Position.initialize_pyramid) with [averaging](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.downscale_local_mean).

examples/demo_dlmbl/python/excercise_1.py
```diff
@@ -0,0 +1,274 @@
...
+    model_config=model_config.copy(),
+    batch_size=BATCH_SIZE,
+    loss_function=torch.nn.functional.mse_loss,
+    schedule="WarmupCosine",
+    log_num_samples=10,
+)
+
+# visualize graph
+model_graph = draw_graph(model, model.example_input_array, depth=2)
+graph = model_graph.visual_graph
```
mattersoflight, line 191: love this graph!

examples/demo_dlmbl/python/excercise_1.py
mattersoflight: Yes! I didn't know that iohub now has support for pyramids too.

examples/demo_dlmbl/python/excercise_1.py
```diff
@@ -0,0 +1,274 @@
...
+graph
+
+# %% [markdown]
+"""
+Configure trainer class.
+Here we use the ``fast_dev_run`` flag to run a sanity check first.
+"""
+
+# %%
+trainer = VSTrainer(accelerator="gpu", fast_dev_run=True)
```
mattersoflight, line 201: Seeing an error on `fry2`
```
/hpc/mydata/shalin.mehta/envs/viscy/lib/python3.10/site-packages/lightning/fabric/plugins/environments/slurm.py:165: PossibleUserWarning: The `srun` command is available on your system but is not used. HINT: If your intention is to run Lightning on SLURM, prepend your python command with `srun` like so: srun python /hpc/mydata/shalin.mehta/envs/viscy/lib/python3.10/s ...
  rank_zero_warn(
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
Running in `fast_dev_run` mode: will run the requested loop using 1 batch(es). Logging and checkpointing is suppressed.


RuntimeError: Lightning can't create new processes if CUDA is already initialized. Did you manually call `torch.cuda.*` functions, have moved the model to the device, or allocated memory on the GPU any other way? Please remove any such calls, or change the selected strategy. You will have to restart the Python kernel.

```

@ziw-liu have you seen this?

examples/demo_dlmbl/python/excercise_1.py
ziw-liu: `batch['index']` is a tuple of image name, time, and z-slice.

examples/demo_dlmbl/python/excercise_1.py
ziw-liu: I didn't see this on Bruno. Maybe a CUDA driver issue?

ziw-liu: Does it persist if you restart the kernel as prompted?

mattersoflight: The error persists. I will try on bruno.

examples/demo_dlmbl/python/excercise_1.py
mattersoflight: The graph's output tensor is 1x1x256x256. Is this the artifact of shape of `example_input_tensor`. Which paramter specifies the number of output channels in model arch?

examples/demo_dlmbl/python/excercise_1.py
ziw-liu: Done in [update pyramid dataset](https://github.com/mehta-lab/viscy/pull/29/commits/f8e2991a5f6cabd70e942872e09059a3b2c0be0a)

examples/demo_dlmbl/python/excercise_1.py
ziw-liu: Sorry this is a bug. Now fixed.

examples/demo_dlmbl/python/excercise_1.py
mattersoflight: Don't see a runtime error on bruno with the same conda env. Please test on AWS nodes after we merge this PR.

ziw-liu: I reproduced the error and traced it to `torchview` copying the model to GPU. I pinned it on CPU in [fix device issue](https://github.com/mehta-lab/viscy/pull/29/commits/84205708b11f4a2df926f125031313edef59a17d). Also fry2 has 4 GPUs, and while that still works by the virtue of lightning, it's unneeded complexity so I'm now pinning it to a single GPU.

examples/demo_dlmbl/python/excercise_1.py
```diff
@@ -0,0 +1,281 @@
...
+
+model = model = VSUNet(
+    model_config=wider_config.copy(),
+    batch_size=BATCH_SIZE,
+    loss_function=torch.nn.functional.mse_loss,
+    schedule="WarmupCosine",
+    log_num_samples=10,
+)
+
+trainer = VSTrainer(
```
mattersoflight, line 249: I think we should write the logs in pwd, such that the folder structure is 04-image-translation/setup.sh, 04-image-translation/viscy, 04-image-translation/lightning_logs, ...

examples/demo_dlmbl/python/excercise_1.py
```diff
@@ -0,0 +1,281 @@
...
+    accelerator="gpu", max_epochs=20, log_every_n_steps=10, default_root_dir=""
+)
+
+trainer.fit(model, datamodule=data_module)
```
mattersoflight, line 253: I am seeing this error: ` PossibleUserWarning: The number of training batches (8) is smaller than the logging interval Trainer(log_every_n_steps=10). et a lower value for log_every_n_steps if you want to see logs for the training epoch.` 

Following the suggestion.

examples/demo_dlmbl/python/excercise_1.py
mattersoflight, line 249: Current logger writes events from two traning runs to the same directory. The tensorboard logs written like this can be confusing for novice.

PR #36, opened 2023-08-18
DLMBL 2023 excercise

ziw-liu:


pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "ipykernel", # used by demo_dlmbl
     "graphviz", # used by demo_dlmbl
     "torchview", # used by demo_dlmbl
+    "pathlib",
```
ziw-liu, line 26: Pathlib is a [standard library](https://docs.python.org/3/library/pathlib.html).
```suggestion
```

mattersoflight: Ah, deleting.

pyproject.toml
```diff
@@ -23,7 +23,8 @@ dependencies = [
     "ipykernel", # used by demo_dlmbl
     "graphviz", # used by demo_dlmbl
     "torchview", # used by demo_dlmbl
-    "pathlib",
+    "tensorboard", # used by demo_dlmbl
```
ziw-liu, line 26:
```suggestion
```
See L17 of this file.

pyproject.toml
```diff
@@ -23,7 +23,8 @@ dependencies = [
...
+    "pillow", # used by demo_dlmbl
```
ziw-liu, line 27:
```suggestion
```
This should come with iohub.

examples/demo_dlmbl/python/solution.py
```diff
@@ -0,0 +1,511 @@
...
+# Part 3: Tune the capacity of networks and other hyperparameters to improve the performance.
+--------------------------------------------------
+
+Learning goals:
+
+- Tweak model hyperparameters, primarily depth and learning rate.
+- Adjust batch size to fully utilize the VRAM
+"""
+
+# %% tags = ["solution"]
```
mattersoflight, line 493: @ziw-liu please prototype and test?

mattersoflight:
@ziw-liu I've tested the whole exercise on dlmbl1 node. In addition to prototyping 3 solutions above, can you suggest how to log all validation samples? At this point, the subsample of validation set comes out empty in multiple runs. 

image

With solutions in hand, I'll write the exercise prompt. Earlier exercise prompts are commented out so that you can just run the script by clicking `Run Above` in the last cell.

examples/demo_dlmbl/python/solution.py
```diff
@@ -0,0 +1,506 @@
...
+    augment = True 
+)
+phase2fluor_data.setup("fit")
+# fast_dev_run runs a single batch of data through the model to check for errors.
+trainer = VSTrainer(accelerator="gpu", devices=[GPU_ID], max_epochs = 40,  fast_dev_run=True)
+
+# trainer class takes the model and the data module as inputs.
+trainer.fit(phase2fluor_model, datamodule=phase2fluor_data)
+
+# %% [markdown]
```
mattersoflight, line 344: @ziw-liu how do you typically log graphs?

ziw-liu: The lightning trainer logs graph automatically when `model.example_input_array` is available. It should be possible to manually do `logger.log_graph(model, array)` as well.

examples/demo_dlmbl/python/solution.py
```diff
@@ -0,0 +1,512 @@
...
+Tips:
+- Set ``default_root_dir`` to store the logs and checkpoints
+in a specific directory.
+"""
+
+# %% tags=["solution"]
+
+GPU_ID = 0
+n_samples = len(phase2fluor_data.train_dataset)
+steps_per_epoch = n_samples // BATCH_SIZE # log data after so many steps.
```
ziw-liu, line 366: This will only log once per epoch. Is that intentional?

ziw-liu: Maybe related to https://github.com/mehta-lab/VisCy/pull/36#issuecomment-1685296228?

ziw-liu:
> can you suggest how to log all validation samples? At this point, the subsample of validation set comes out empty in multiple runs.

I think I know what's happening here. The validation samples are the first sample of the first `log_num_samples` batches. So if total validation steps is 1 it will only have 1 sample. I can modify the behavior to only use the first batch in the epoch, with the caveat that `log_num_samples` will then have to be smaller than batch size.

examples/demo_dlmbl/python/solution.py
```diff
@@ -0,0 +1,503 @@
...
+    "out_channels": 2,
+    "residual": True,
+    "dropout": 0.1, # dropout randomly turns off weights to avoid overfitting of the model to data.
+    "task": "reg", # reg = regression task.
+}
+
+phase2fluor_model = VSUNet(
+    model_config=phase2fluor_config.copy(),
+    batch_size=BATCH_SIZE,
+    loss_function=torch.nn.functional.l1_loss,
```
ziw-liu, line 321: Is MAE better than MSE here?

examples/demo_dlmbl/python/solution.py
ziw-liu: > It should be possible to manually do `logger.log_graph(model, array)` as well.

It won't work here though, as logging is completely disabled when `fast_dev_run=True`.

examples/demo_dlmbl/python/solution.py
mattersoflight: Yes, I meant to log the losses and image samples once per epoch.

examples/demo_dlmbl/python/solution.py
mattersoflight: trying l1_loss

mattersoflight: both look similar for this data.

examples/demo_dlmbl/python/solution.py
ziw-liu: https://github.com/mehta-lab/VisCy/pull/36/commits/9d29413a3b7e28982d093db2c0f9917a4445e83f seems to fix this.

mattersoflight:
@ziw-liu the conflicts in engine.py seem to be related to how you are logging the loss in two branches. Once you resolve them. Feel free to merge this branch into #37. I can start a smaller PR with updates from the course. 

```
<<<<<<< dlmbl2023
        self.log("loss/validate", loss, batch_size=self.batch_size, sync_dist=True)
        if batch_idx == 0:
            self.validation_step_outputs.extend(
=======
        self.log("loss/validate", loss, sync_dist=True)
        if batch_idx < self.log_num_samples:
            self.validation_step_outputs.append(
>>>>>>> 21d-upscale-decoder
                self._detach_sample((source, target, pred))
            )
```

Issue #43, opened 2023-08-30
HCSDataModule not letting me overwrite when fitting


edyoshikun:
I was expecting that if you call  `HCSDataModule().setup('fit')` the DataModule should fit the data and re-write the normalization dictionary. However, when this is called twice in a row, we get:

```bash
KeyError                                  Traceback (most recent call last)
[/home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py](https://vscode-remote+ssh-002dremote-002bec2-002d3-002d144-002d39-002d9-002eus-002deast-002d2-002ecompute-002eamazonaws-002ecom.vscode-resource.vscode-cdn.net/home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py) in line 14
      [34](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=33) # %%
      [35](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=34) data_module = HCSDataModule(
      [36](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=35)     input_data_path,
      [37](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=36)     source_channel="Phase3D",
   (...)
     [45](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=44)     augment=False,  # Turn off augmentation for now.
     [46](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=45) )
---> [47](file:///home/eduardoh/vs_data/1-test_dataloader/edhirata_dataloader.py?line=46) data_module.setup("fit")

File [~/VisCy/viscy/light/data.py:404](https://vscode-remote+ssh-002dremote-002bec2-002d3-002d144-002d39-002d9-002eus-002deast-002d2-002ecompute-002eamazonaws-002ecom.vscode-resource.vscode-cdn.net/home/eduardoh/VisCy/~/VisCy/viscy/light/data.py:404), in HCSDataModule.setup(self, stage)
    [402](file:///home/eduardoh/VisCy/viscy/light/data.py?line=401) dataset_settings = dict(channels=channels, z_window_size=self.z_window_size)
    [403](file:///home/eduardoh/VisCy/viscy/light/data.py?line=402) if stage in ("fit", "validate"):
[...]

PR #45, opened 2023-08-31
Example workflow

ziw-liu:
Adds a short description of the virtual staining workflow in its current state.

### DIFF
--- a/examples/configs/test_example.yml
+++ b/examples/configs/test_example.yml
@@ -1,0 +1,73 @@
+# lightning.pytorch==2.0.4
+seed_everything: true
+trainer:
+  accelerator: auto
+  strategy: auto
+  devices: auto
+  num_nodes: 1
+  precision: 32-true
+  callbacks: null
+  fast_dev_run: false
+  max_epochs: null
+  min_epochs: null
+  max_steps: -1
+  min_steps: null
+  max_time: null
+  limit_train_batches: null
+  limit_val_batches: null
+  limit_test_batches: null
+  limit_predict_batches: null
+  overfit_batches: 0.0
+  val_check_interval: null
+  check_val_every_n_epoch: 1
+  num_sanity_val_steps: null
+  log_every_n_steps: null
+  enable_checkpointing: null
+  enable_progress_bar: null
+  enable_model_summary: null
+  accumulate_grad_batches: 1
+  gradient_clip_val: null
+  gradient_clip_algorithm: null
+  deterministic: null
+  benchmark: null
+  inference_mode: true
+  use_distributed_sampler: true
+  profiler: null
+  detect_anomaly: false
+  barebones: false
+  plugins: null
+  sync_batchnorm: false
+  reload_dataloaders_every_n_epochs: 0
+  default_root_dir: null
+model:
+  architecture: null
+  model_config: {}
+  loss_function: null
+  lr: 0.001
+  schedule: Constant
+  log_num_samples: 8
+  test_cellpose_model_path: null
+  test_cellpose_diameter: null
+  test_evaluate_cellpose: false
+data:
+  data_path: null
+  source_channel: null
+  target_channel: null
+  z_window_size: null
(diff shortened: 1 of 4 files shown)


mattersoflight (approved):
LGTM

mattersoflight:
This addresses #43.

examples/configs/test_example.yml
```diff
@@ -0,0 +1,73 @@
...
+  barebones: false
+  plugins: null
+  sync_batchnorm: false
+  reload_dataloaders_every_n_epochs: 0
+  default_root_dir: null
+model:
+  architecture: null
+  model_config: {}
+  loss_function: null
+  lr: 0.001
```
edyoshikun, line 46: At least from the examples at the DL@MBL course 1e-4 worked better and I think that is the default in most networks, but if this typically works better in our networks then great.

examples/configs/test_example.yml
```diff
@@ -0,0 +1,73 @@
...
+  - 256
+  - 256
+  augment: true
+  caching: false
+  normalize_source: false
+  ground_truth_masks: null
+  train_z_scale_range:
+  - 0.0
+  - 0.0
+  train_noise_std: 0.0
```
edyoshikun, line 70: I thought this were obtained from the HCSDataModule.setup('fit')(?)

examples/configs/test_example.yml
ziw-liu: The best LR is highly data and model dependent, and there is not a good default that works well for everything. This LR works for 2.5D UNet with 64 patches per batch. For 2.5D LUNeXt the gradient will explode after a few epochs so I've been using 2e-4.

examples/configs/test_example.yml
ziw-liu: It infers the correct shape based on the user input here:

https://github.com/mehta-lab/VisCy/blob/3f46b3b8e210c478ecceb7b1ec05618c2abb045f/viscy/light/data.py#L440-L448

Can you explain more?

edyoshikun:
LGTM! Thanks @ziw-liu

edyoshikun (approved):

examples/configs/test_example.yml
edyoshikun: Sorry, I had the wrong intuition on what these were supposed to do. Since I didn't know we could run the config through the CLI.

edyoshikun:
Maybe this is for a separate issue, but I had also trouble running the metrics/test command. I think this was just a lack of time on my side during the course to fully explore pytorch lightning framework. I think that if we run it together @ziw-liu  this comment can be solved and we can add further documentation as separate PR.

Soorya19Pradeep:
Thank for putting together this example @ziw-liu! It was very easy to navigate and documentation was great!
I was able to navigate through the solution and train a 2D HEK nucleus-membrane prediction model using uninfected cells in the TICM0036 dataset acquired on Dragonfly microscope.  I was able to test the model by performing prediction on OC43-infected HEK cells at 48hpi. 

I did make changes on the line:
https://github.com/mehta-lab/VisCy/blob/c13176d8d9b21cfc3a9b2ead913f748bc0dd775d/examples/demo_dlmbl/solution.py#L396

I changed this based on @ziw-liu 's suggestion.
```
phase2fluor_config = {
    "num_filters": [24, 48, 96, 192, 384],
    "in_channels": 1,
    "out_channels": 2,
    "residual": True,
    "dropout": 0.1,  # dropout randomly turns off weights to avoid overfitting of the model to data.
    "task": "reg",  # reg = regression task.
}

phase2fluor_model = VSUNet(
    architecture="2D",
    model_config=phase2fluor_config.copy(),
    loss_function=torch.nn.functional.l1_loss,
    schedule="WarmupCosine",
    log_num_samples=5,  # Number of samples from each batch to log to tensorboard.
    example_input_yx_shape=YX_PATCH_SIZE,
)
```

PR #49, opened 2023-09-06
Configure the number of  image samples logged at each epoch and batch

### LINKED ISSUE #46: Configure sample image logging during training
For 2D models, logging one sample from each batch resulted in too few samples being logged for the validation set, since the number of FOVs for validation is relatively small with regard to typical batch sizes. However, changing the logging scheme to use the first N samples from only the first batch will log samples from only the first FOV in validation due to sliding window sampling. To support both use cases, I propose the following change to the logging interface:

Current:

```yaml
model:
  log_num_samples: 12
```

Change to:

```yaml
model:
  log_batches_per_epoch: 4
  log_samples_per_batch: 3
```

So for 2D training, these numbers can be (1, 12) so that even the validation epoch only has 1 batch, this will still log the same number of images.

ziw-liu:
Resolves #46.

### DIFF
--- a/examples/configs/fit_example.yml
+++ b/examples/configs/fit_example.yml
@@ -44,7 +44,8 @@ model:
   loss_function: null
   lr: 0.001
   schedule: Constant
-  log_num_samples: 8
+  log_batches_per_epoch: 8
+  log_samples_per_batch: 1
 data:
   data_path: null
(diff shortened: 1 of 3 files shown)


edyoshikun (approved):
LGTM! thanks @ziw-liu

PR #51, opened 2023-09-07
Fix center slice metrics for 3D output

ziw-liu:
Fix the center slice index for 3D output when computing metrics.


edyoshikun (approved):
thanks @ziw-liu . This should fix the issue I was having running the test.

PR #57, opened 2023-09-28
Fix predicting new channels in an existing store

ziw-liu:
Open `HCSPredictionWriter.plate` after appending new channels so that the new channels are visible when retrieving channel indices.


edyoshikun (approved):
Thanks!

PR #58, opened 2023-10-05
Baseline 3D-LUNeXt

ziw-liu:
Baseline upgrade of the 3D-LUNeXt model.


mattersoflight (approved):

ziw-liu:
@edyoshikun I'm trying to keep this the same as we are using for the mantis figures. I've tested that `/hpc/projects/comp.micro/mantis/2023_08_18_A549_DRAQ5/3-registration/prediction/predict.yml` still predicts the same results.

I will handle renaming and other refactoring after this merges.

PR #59, opened 2023-11-03
Preprocess CLI and source scaling during prediction

ziw-liu:
Moved the separate preprocessing script into the lightning CLI as the `viscy preprocess` subcommand.

The command only has 1 required argument (path) and 3 optional ones, so using config files will be largely optional. For most cases something like this will suffice:

```sh
viscy preprocess --data_path /path/to/data.zarr --num_workers 16
```

In addition to plate metadata, the preprocessing step will now also write the dataset statistics to each FOV (resulting in them having both plate and FOV level statistics), so that prediction can also be done with a single FOV in a plate as input. (Note that it does *not* support wildcards.) This allows for parallel processing. However a manual collecting step is needed since this is not a concurrent design.

Also added an option (`data.predict_scale_source: float | None`) to scale (multiply) the input channels by a factor during the prediction stage.


viscy/light/data.py
```diff
@@ -344,13 +351,14 @@ def __init__(
         augment: bool = True,
         caching: bool = False,
         normalize_source: bool = False,
-        ground_truth_masks: str = None,
+        ground_truth_masks: Optional[Path] = None,
         train_z_scale_range: tuple[float, float] = [0, 0],
         train_noise_std: float = 0.0,
         train_patches_per_stack: int = 1,
+        predict_scale_source: Optional[float] = None,
```
edyoshikun, line 358: Here how do we select which source channel gets the scaling or does it apply it to both?

viscy/light/data.py
```diff
@@ -318,7 +320,10 @@ class HCSDataModule(LightningDataModule):
     :param bool caching: whether to decompress all the images and cache the result,
         will store in ``/tmp/$SLURM_JOB_ID/`` if available,
         defaults to False
-    :param str ground_truth_masks: path to the ground truth segmentation masks,
+    :param bool normalize_source: whether to normalize the source channel,
```
edyoshikun, line 323: source channels?

viscy/light/data.py
edyoshikun: Never mind we set on the config file which channels are source and target. Thanks

edyoshikun (commented):
Thanks @ziw-liu this simplifies the preprocessing. I still need to test the scaling, but if this needs to be merged to main we can do so.

viscy/light/data.py
ziw-liu: If this is set to `True` then all source channels will be normalized.

ziw-liu:
> I still need to test the scaling, but if this needs to be merged to main we can do so.

I didn't see strong benefits from scaling. Since this PR is mainly for your slurmkit workflow it's your call whether we should merge as-is.

edyoshikun (approved):
ready to merge! @ziw-liu

PR #60, opened 2023-11-06
Example script for parsing CLI config files in Python

ziw-liu:
Shows how to subclass the trainer with a placeholder method so the CLI parser can be reused.

Adapted from https://github.com/Lightning-AI/lightning/issues/12302#issuecomment-1110425635.

What we actually want is probably https://github.com/Lightning-AI/lightning/pull/18105.


ziw-liu:
https://github.com/Lightning-AI/pytorch-lightning/pull/18105 is now available in lightning 2.3.0.

PR #61, opened 2023-11-21
Configurable augmentations

ziw-liu:
Made training-time augmentations configurable through YAML config files and CLI arguments.

One caveat is that the MONAI transforms has to be redefined so that the type hints in their `__init__()` signatures are supported by `jsonargparse`. This adds some boilerplate code in the library.


ziw-liu:
Removed auto-generated notebooks since they are difficult to maintain and are duplicates of the python files.

examples/configs/fit_example.yml
```diff
@@ -47,17 +26,52 @@ model:
...
+  target_channel:
+    - target_1
+    - target_2
+  z_window_size: 5
+  split_ratio: 0.8
+  batch_size: 32
+  num_workers: 16
+  yx_patch_size: [256, 256]
+  augmentations:
+    - class_path: monai.transforms.RandWeightedCropd
```
edyoshikun, line 40: change the monai to viscy

edyoshikun (approved):
other than my comment. this LGTM! I also tested this with the newest mantis data model.

PR #66, opened 2024-01-25
checkpoint as a model config parameter for warmup cosine learning rates

edyoshikun:
This PR adds the an extra parameter for the CLI to provide the checkpoint path to be loaded such that is compatible when using warmup cosine learning rates. By default, the lightning CLI has a `ckpt_path`, but this loads the previous checkpoint's learning rate as well as the weights. When using the latter option, the re-training for next model starts as a 'sine' function rather than cosine.

### DIFF
--- a/viscy/light/engine.py
+++ b/viscy/light/engine.py
@@ -122,7 +122,7 @@ class VSUNet(LightningModule):
         loss_function: Union[nn.Module, MixedLoss] = None,
         lr: float = 1e-3,
         schedule: Literal["WarmupCosine", "Constant"] = "Constant",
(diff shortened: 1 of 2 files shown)


viscy/light/engine.py
```diff
@@ -144,6 +146,7 @@ def __init__(
         self.schedule = schedule
         self.log_batches_per_epoch = log_batches_per_epoch
         self.log_samples_per_batch = log_samples_per_batch
+
```
ziw-liu, line 149:
```suggestion
```

ziw-liu (approved):

PR #67, opened 2024-02-08
Masked autoencoder pre-training for virtual staining models

ziw-liu:


edyoshikun (approved):
This branch has worked well for both 2.2D and 3D LUNeXT training/test so I am happy with it. 
- Combined and Concatenated data loader work
- Augmentations show no issue

mattersoflight:
@ziw-liu Looks like this is now the de facto branch for both training recipes: end-to-end training and pre-training + fine-tuning. If it also provides a reasonably stable implementation of concatenated data loaders, please merge in `main`.

ziw-liu:
> Looks like this is now the de facto branch for both training recipes: end-to-end training and pre-training + fine-tuning. Can you confirm if that is the case? Please merge in main if the merge won't break the current experiments.

This is the case. However note that I didn't do comprehensive backwards compatibility testing so it ***could*** break previously trained models.

PR #68, opened 2024-02-24
configurable normalizations

edyoshikun:
This PR adds configurable normalizations for both the source and targets similar to the transformations/augmentations that are applied to the samples.


examples/configs/fit_example.yml
```diff
@@ -36,6 +36,19 @@ data:
   batch_size: 32
   num_workers: 16
   yx_patch_size: [256, 256]
+  normalizations:
+  - class_path: viscy.transforms.NormalizeSampled
```
ziw-liu, line 40: Need indentation here.

viscy/utils/meta_utils.py
```diff
@@ -127,6 +127,7 @@ def generate_normalization_metadata(
                 process=f"calculating channel statistics {channel}/{list(channel_ids)}",
             )
             position_statistics = dataset_statistics | {
+                "dataset_statistics": dataset_level_statistics,
```
ziw-liu, line 130: It should already exist. Can you test with a toy dataset without this line?

https://github.com/mehta-lab/VisCy/blob/529340db3d2750a0a5b2eb04a2a452df32e3d56f/viscy/utils/meta_utils.py#L112-L114

ziw-liu: I suspect that the dataset we looked at earlier was processed with an older version.

ziw-liu: See https://github.com/mehta-lab/VisCy/pull/68/commits/0a15f311453565fa6640c1e3deaa6d7ceadeb008 and https://github.com/mehta-lab/VisCy/pull/68/commits/c13400c0c951e9130ab70e9a255ff97f98676e68

edyoshikun:
@ziw-liu just tested with new dataset using the `preprocessing` and we can do the normalizations

ziw-liu (approved):

PR #71, opened 2024-03-04
2D FCMAE

ziw-liu:
Add option to not perform convolution in the head of the FCMAE models.

Fixes for multi-dataset trainer.

Fix default when computing metrics on images with very high instance counts.


viscy/light/engine.py
```diff
@@ -305,8 +323,16 @@ def on_train_epoch_end(self):
         self.training_step_outputs = []
 
     def on_validation_epoch_end(self):
+        super().on_validation_epoch_end()
         self._log_samples("val_samples", self.validation_step_outputs)
         self.validation_step_outputs = []
+        # average within each dataloader
```
edyoshikun, line 329: Here I know that this is the end of the validation meaning that it won't backprop and therefore maybe no need to detach the tensor before doing any logging. Is this the common practice? I know that for the `train_step` and `validation_step` is more relevant and important If we don't detach here it doesnt affect it any way? Just curious..

ziw-liu: To log the loss value (through lightning logger), the detaching is automatic. We only need to care about it when logging manually (images).

edyoshikun (approved):
I tested doing the following without issues:
- 2D predictions using a549 2D data
- 3D predictions using mantis A549 data

I don't expect it to break anything related to the dataloaders since I was using the hcs.py. I like the `prepare_data_per_node` flags for caching. I will just need to update my scripts to add this flag from now on. 

Thank you so much for this Ziwen!

PR #74, opened 2024-04-15
Filter empty detections in labels

ziw-liu:
Fix overcounting instances when the label values are not contiguous.


edyoshikun (approved):
This seems to work for when the labels are not in a sequence. I tested this with the datasets that I got and compared the mAP with PQ's RQ and get very similar results.

PR #76, opened 2024-05-09
solution notebook

esgomezm:


mattersoflight:
This will be resolved when #100 ie merged.

PR #80, opened 2024-05-25
Fix 3D to 2D prediction with UNeXt2 model

### LINKED ISSUE #78: Three channel output using covnext model
I want to train the covnext model for the infection classifier problem. The model will use two or more input channels (phase + HSP90 channels + infection score channels, etc.) and should output three channels (background + uninfected + infected channels). Currently, I can perform this training using 2D and 2.5D unets, but not using covnext. It doesn't allow three-channel output due to a hardcoded scaling parameter value. @ziw-liu, can you help with this? Thanks!

ziw-liu:
Removed the "2.1D" architecture and made the output stack depth of "2.2D" configurable between 2D and 3D.

Fix #78.


ziw-liu:
@Soorya19Pradeep see the updated example script for the configuration you requested.

mattersoflight (approved):

PR #85, opened 2024-06-11
Add preprint reference to README

ziw-liu:
Add the robust virtual staining preprint to README, along with some minor edits.


edyoshikun (commented):
do we know how were the previous architecture models generated? it might be helpfult to update that image as well.

edyoshikun (approved):
Otherwise this LGTM!

ziw-liu:
> do we know how were the previous architecture models generated? it might be helpfult to update that image as well.

I think they were draw by hand. (I took them from the eLife 2020 paper).

PR #86, opened 2024-06-12
Fix architecture name in network diagram script

ziw-liu:


edyoshikun (approved):
I really like the new diagrams. Do you know if these can be rotated? Perhaps this can be used to replace the Readme.md networks.
 
Works great! Thanks

ziw-liu:
> I really like the new diagrams. Do you know if these can be rotated? Perhaps this can be used to replace the Readme.md networks.
> 
> Works great! Thanks

They can be rotated by `graph_dir="LR"`. But they become extremely wide since all the shape text is now in one line.

PR #88, opened 2024-06-12
Test on Python 3.12

ziw-liu:


edyoshikun (approved):
For the 3rd time... thanks again! I'll fix whatever is wrong with my git. 
I think we missed the monai bump, so I added this here. 
I tested this with the new branch, the installation, and prediction.

PR #89, opened 2024-06-13
Add the scale metadata to the output_stores

edyoshikun:
PR adds the scale metadata to the output stores. 
~This is solution requires one to add an input_metadata_store that has the proper scale.~

### DIFF
--- a/viscy/light/predict_writer.py
+++ b/viscy/light/predict_writer.py
@@ -48,8 +48,8 @@ def _blend_in(old_stack: NDArray, new_stack: NDArray, z_slice: slice) -> None:
 class HCSPredictionWriter(BasePredictionWriter):
     """Callback to store virtual staining predictions as HCS OME-Zarr.
 
-    :param str metadata_store: Path to the zarr store input
     :param str output_store: Path to the zarr store to store output
+    :param str metadata_store: Path to the OME-Zarr dataset to copy scale metadata from
     :param bool write_input: Write the source and target channels too
         (must be writing to a new store),
         defaults to False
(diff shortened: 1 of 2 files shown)


viscy/light/predict_writer.py
```diff
@@ -57,13 +59,33 @@ class HCSPredictionWriter(BasePredictionWriter):
 
     def __init__(
         self,
+        metadata_store: str,
```
ziw-liu, line 62: Should be `str | None = None`?

viscy/light/predict_writer.py
```diff
@@ -57,13 +59,33 @@ class HCSPredictionWriter(BasePredictionWriter):
...
+        # Update the scale metadata
+        try:
+            with open_ome_zarr(metadata_store, mode="r") as meta_plate:
+                for _, pos in meta_plate.positions():
+                    self._dataset_scale = [
+                        TransformationMeta(type="scale", scale=pos.scale)
+                    ]
+                    _logger.debug(f"Dataset scale {pos.scale}.")
+                    break
+        except IOError:
```
ziw-liu, line 84: When is this triggered?

edyoshikun: I'll swap this since we changed the input `metadata_store=None` as default behaviour. This was getting triggered if you passed anything other than a zarr.

ziw-liu:
@edyoshikun I forgot that there _is_ a runtime-only reference of the datamodule from the trainer... So I removed the API change. Can you test and update OP?

edyoshikun (commented):
This LGTM! it also worked. Thanks I think it is definetly a cleaner way to handle it.

ziw-liu (approved):

PR #93, opened 2024-06-20
Scale metadata handling for positions

edyoshikun:
This PR fixes the bug for the scale_metadata to retrieve the scale metadata if the input is a Positions or Plate level.


ziw-liu (approved):
LGTM!

PR #112, opened 2024-07-19
Update demo setup script

ziw-liu:


ziw-liu:
I pinned the version to v0.2.0 stable, which isn't released yet. It can be released after merging this PR.

edyoshikun (approved):

PR #122, opened 2024-08-02
Migrate from wandb to tensorboard

### LINKED ISSUE #116: switching to tensorboard logger for organelle phenotyping
Hi @alishbaimran @ziw-liu I looked carefully into model logging and profiling yesterday and figured that W&B is not the optimal backend for our projects. Please start using [tensorboard logger](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.tensorboard.html) in your [training script](https://github.com/mehta-lab/VisCy/blob/infection_state/viscy/applications/contrastive_phenotyping/training_script.py) on `infection_state` branch. 

If this turns out to be tricky,  I will start a PR after Tuesday to help with the transition.

Here are the features we need for the project, and options. 
--------

**Model visualization:**  There doesn’t seem to be a way to extract[ images from W&B logs to illustrate training runs](https://github.com/wandb/wandb/issues/5641). See [this movie](https://www.biorxiv.org/content/biorxiv/early/2024/06/23/2024.05.31.596901/DC6/embed/media-6.mp4?download=true) for the visualizations we would need to interpret models. We have code to make movies like this tensorboard and W&B doesn’t yet provide an API to do this.

**Profiling resources:** W&B’s visualization of resources is useful, but resource usage can be profiled with lightning's [SimpleProfiler](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.profilers.SimpleProfiler.html#lightning.pytorch.profilers.SimpleProfiler) and [PytorchProfiler](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.profilers.PyTorchProfiler.html#lightning.pytorch.profilers.PyTorchProfiler). The [PytorchProfiler](https://pytorch.org/tutorials/_static/img/profiler_overview1.png) is more feature complete than W&B.

**Profiling gradients:** W&B makes it particularly easy to visualize vanishing gradients in a layer, but this can also be done by logging `.grad` property of tensors to tensorboard.  We should look for a convenient API that plays well with lightning. @ziw-liu have you come across any good solutions? 

**Sharing logs:** W&B makes it easy to share logs via URL, but it is only marginally harder to use tensorboard to share logs by pointing to a directory. 

Lastly, **monetary and time cost:** For the active team of 4, W&B pricing structure will cost us the cost of attendance at a conference for one team member over 6-9 months. In addition to $50/user/month, we’d pay $3/100GB/month.  If we use W&B for this project, we will fragment virtual staining and organelle phenotyping logs in different systems. If we use W&B for all of our projects, it is expensive.

ziw-liu:
Fix #116.

Example logs:

```sh
/hpc/projects/intracellular_dashboard/viral-sensor/infection_classification/models/test_tb
```

### DIFF
--- a/applications/contrastive_phenotyping/demo_fit.py
+++ b/applications/contrastive_phenotyping/demo_fit.py
@@ -1,6 +1,8 @@
 from lightning.pytorch import Trainer
 from lightning.pytorch.callbacks import ModelCheckpoint
 from lightning.pytorch.loggers import TensorBoardLogger
+from lightning.pytorch.callbacks import DeviceStatsMonitor
+
 
 from viscy.data.triplet import TripletDataModule
 from viscy.light.engine import ContrastiveModule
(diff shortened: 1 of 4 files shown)


mattersoflight:
@ziw-liu thanks for these improvements! I read through the code and tried `demo_fit.py`. Training of losses, metrics, and images looks good. But, the model graph was not logged. Please add the code to do so.

mattersoflight (approved):
works well

PR #124, opened 2024-08-05
Merge updated training w/ tb logging

alishbaimran:
Updated training script to include tensorboard logging


viscy/light/engine.py
```diff
@@ -723,12 +728,18 @@ def validation_step(
...
 
     def on_validation_epoch_end(self) -> None:
         super().on_validation_epoch_end()
+        val_loss_epoch = torch.stack(self.validation_losses).mean()
+        self.log(
+            "val/loss_epoch", val_loss_epoch, prog_bar=True, logger=True, sync_dist=True
+        )
         self._log_samples("val_samples", self.validation_step_outputs)
         self.validation_step_outputs = []
+        self.validation_losses = []
```
ziw-liu, line 742: Logging loss per epoch is already done automatically. There is no need to do this manual aggregation.

viscy/light/engine.py
```diff
@@ -40,6 +40,10 @@
 except ImportError:
     CellposeModel = None
 
+try:
+    import wandb
+except ImportError:
+    wandb = None
```
ziw-liu, line 46: Why?

viscy/data/hcs.py
```diff
@@ -26,6 +27,11 @@
 
 from viscy.data.typing import ChannelMap, DictTransform, HCSStackIndex, NormMeta, Sample
 
+warnings.filterwarnings(
+    "ignore",
+    category=UserWarning,
+    message="To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).",
+)
```
ziw-liu, line 34: If there is a warning, it should be fixed instead of suppressed. Also I believe I already fixed this.

viscy/light/engine.py
```diff
@@ -798,4 +809,4 @@ def on_predict_epoch_end(self) -> None:
             }
         )
 
-        df.to_csv("embeddings2/multi_resnet_predicted_metadata.csv", index=False)
+        df.to_csv("embeddings4/1_multi_resnet_predicted_metadata.csv", index=False)
```
ziw-liu, line 812: This should be configurable. No change to library code should be needed just for another training run.

viscy/representation/contrastive.py
```diff
@@ -86,9 +86,11 @@ def __init__(
     def forward(self, x):
         x = self.stem(x)
         embedding = self.encoder(x)
+        embedding_norm = F.normalize(embedding, p=2, dim=1)
+        print("embedding shape", embedding_norm.shape)
```
ziw-liu, line 90:
```suggestion
```

ziw-liu (changes requested):
@alishbaimran I highly encourage you to examine the diff and make sure that they are actually intended behavior. Also it is confusing to me why https://github.com/mehta-lab/VisCy/commit/0ba38c71c179bcf26c0cf33a0cee0facb19ca2f6 does not show up in the history of this branch, but instead squashed into faaabd5f8a975c8e5737e368716c227c71452788. This happens with all your branches and is not a clean way to manage code.

mattersoflight:
@alishbaimran here is a commit graph to emphasize Ziwen's point. 
image

Since the commits of changes we made and tested doesn't show up in your branch, we have to examine your files very closely for review. 

Look at this carefully with Ziwen tomorrow and figure out how you can avoid this. 


We follow this practice for creating branches, merging them, and deleting them: https://docs.github.com/en/get-started/using-github/github-flow

mattersoflight:
@alishbaimran here is a concrete example of how your workflow can overwrite an important change. 

2 weeks ago I cleaned up and tested `graphs_ConvNeXt_ResNet.py` to test the model construction.

image

That edit is now not in your latest commit and I am getting an import error

image

I will closely examine this PR, but be mindful of following the GitHub flow process to avoid this in the future.

mattersoflight:
I fixed one file. 

@ziw-liu @alishbaimran which branch are you using currently?

viscy/representation/contrastive.py
```diff
@@ -86,9 +86,11 @@ def __init__(
     def forward(self, x):
         x = self.stem(x)
         embedding = self.encoder(x)
+        embedding_norm = F.normalize(embedding, p=2, dim=1)
```
ziw-liu, line 89: Instead of returning normalized embedding here, it can return the embeddings and let the caller do normalization, which is parameter-free. This allows the caller to have access to both, as normalization is destructive.

ziw-liu:
> which branch are you using currently to train models

I'm branching off of main: https://github.com/mehta-lab/VisCy/pull/126.

PR #128, opened 2024-08-08
version lighting CLI example

mattersoflight:
Documenting how to use lighting CLI for contrastive learning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ __pycache__/
 # written by setuptools_scm
 */_version.py
 
+# slurm output files
+slurm-*
+
 # Distribution / packaging
 .Python
 build/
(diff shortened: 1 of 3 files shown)


mattersoflight:
@ziw-liu  why do you copy the zarr store to `/tmp` and then delete it? Is this i/o expense needed if we use VAST?

ziw-liu:
> @ziw-liu which *.zarr are copied during training and why is this cleanup needed?
> 
> ```
> function cleanup() {
>   rm -rf /tmp/$SLURM_JOB_ID/*.zarr
>   echo "Cleanup Completed."
> }
> 
> trap cleanup EXIT
> ```

The dataset is copied to tmpfs if `data.caching: true`. I have it in my job script in case I use caching. Otherwise it doesn't do anything.

ziw-liu (approved):

PR #129, opened 2024-08-08
Adding support for pyramids

edyoshikun:
This PR adds the support for pyramids

### DIFF
--- a/viscy/data/hcs.py
+++ b/viscy/data/hcs.py
@@ -104,6 +104,8 @@ class SlidingWindowDataset(Dataset):
     :param ChannelMap channels: source and target channel names,
         e.g. ``{'source': 'Phase', 'target': ['Nuclei', 'Membrane']}``
     :param int z_window_size: Z window size of the 2.5D U-Net, 1 for 2D
+    :param int pyramid_resolution: pyramid level.
+        defaults to 0 (full resolution)
     :param DictTransform | None transform:
         a callable that transforms data, defaults to None
     """
@@ -113,16 +115,21 @@ class SlidingWindowDataset(Dataset):
         positions: list[Position],
         channels: ChannelMap,
         z_window_size: int,
+        pyramid_resolution: int = 0,
         transform: DictTransform | None = None,
     ) -> None:
         super().__init__()
         self.positions = positions
         self.channels = {k: _ensure_channel_list(v) for k, v in channels.items()}
         self.source_ch_idx = [
-            positions[0].get_channel_index(c) for c in channels["source"]
+            positions[pyramid_resolution].get_channel_index(c)
+            for c in channels["source"]
         ]
         self.target_ch_idx = (
-            [positions[0].get_channel_index(c) for c in channels["target"]]
+            [
+                positions[pyramid_resolution].get_channel_index(c)
+                for c in channels["target"]
+            ]
             if "target" in channels
             else None
         )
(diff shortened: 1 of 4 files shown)


viscy/data/hcs.py
```diff
@@ -104,6 +104,8 @@ class SlidingWindowDataset(Dataset):
     :param ChannelMap channels: source and target channel names,
         e.g. ``{'source': 'Phase', 'target': ['Nuclei', 'Membrane']}``
     :param int z_window_size: Z window size of the 2.5D U-Net, 1 for 2D
+    :param int pyramid_resolution: pyramid level.
```
ziw-liu, line 107: Outdated type hint. Also I'm wondering if this is the best name, since it's valid for any multi-array FOV group that may or may not be a pyramid.

mattersoflight:
@edyoshikun thanks - this implementation seems backward compatible with previous configs for training and prediction for virtual staining and contrastive learning tasks. Can you check with the configs we are sharing with the preprint if they need any change (e.g., explicitly setting the scale = 0)?

viscy/data/hcs.py
edyoshikun: Do you have any name suggestions? I can add in the docstring that "0" is (highest-resolution). 
This parameter is also totally optional. It's defaulted to `"0"`, and only if you have a multiscale you would probably look for this parameter.

edyoshikun:
@mattersoflight we don't need to change the previous configs because by default the Dataloader will be set to "0" the highest resolution if the parameter is not provided in the config. 

For the preprint, I thought we were pointing them to `viscy=0.2.0`, which may or may not have this PR. If we are adding this PR as part of 0.2.0, I'll make the changes.

talonchandler:
I've tested the basic `HCSDataModule` with multiple resolution levels, and it's working well. I can load random patches at any resolution level in my pyramid. Thanks @edyoshikun. 

@edyoshikun I'm happy to leave this as a branch while we test this out end to end during the course.

viscy/data/hcs.py
ziw-liu: Yes for now explaining in docstring is sufficient.

viscy/data/hcs.py
```diff
@@ -104,6 +104,8 @@ class SlidingWindowDataset(Dataset):
     :param ChannelMap channels: source and target channel names,
         e.g. ``{'source': 'Phase', 'target': ['Nuclei', 'Membrane']}``
     :param int z_window_size: Z window size of the 2.5D U-Net, 1 for 2D
+    :param str pyramid_resolution: pyramid level.
+        defaults to 0 (full resolution)
```
ziw-liu, line 108:
```suggestion
        defaults to "0" (full resolution)
```

ziw-liu:
The code changes look good, but we should test that training actually works before merging.

viscy/data/hcs.py
```diff
@@ -219,7 +223,7 @@ def __getitem__(self, index: int) -> Sample:
         sample = {
             "index": sample_index,
             "source": self._stack_channels(sample_images, "source"),
-            "norm_meta": norm_meta,
+            # "norm_meta": norm_meta,
```
ziw-liu, line 226: What is this for?

viscy/data/hcs.py
```diff
@@ -318,6 +324,7 @@ def __init__(
         augmentations: list[MapTransform] = [],
         caching: bool = False,
         ground_truth_masks: Path | None = None,
+        pyramid_resolution: str = "0",
```
ziw-liu, line 327: #218 used `array_key` for this parameter, which is a more general naming:

https://github.com/mehta-lab/VisCy/blob/1d3261d1e592e1a0c518670295dc6c84ee0806fb/viscy/data/mmap_cache.py#L40

ziw-liu (changes requested):
While this might work for training, the prediction writer will get a shape mismatch error and needs to be patched too.

PR #131, opened 2024-08-08
Configurable drop path rate in contrastive models

ziw-liu:
Make `drop_path_rate` configurable in `ContrastiveEncoder` and `ContrastiveModule`. This value was previously hard-coded to be `0.2` for all models (ConvNeXt and ResNet), while `timm`'s default is `0.0`.


ziw-liu:
Rebased on `main` (including tokenizing stem).

mattersoflight (approved):
LGTM!

PR #132, opened 2024-08-10
Config-based prediction with Xarray-based output format

ziw-liu:
Add CLI to predict embeddings.

Use Xarray to organize and store the predictions so that specific conditions and cells can be indexed directly.


applications/contrastive_phenotyping/contrastive_cli/predict.yml
```diff
@@ -0,0 +1,44 @@
...
+  callbacks:
+    - class_path: viscy.light.embedding_writer.EmbeddingWriter
+      init_args:
+        output_path: 2024_06_13-tokenized-drop_path_0_0.zarr
+  inference_mode: true
+model:
+  backbone: convnext_tiny
+  in_channels: 2
+  in_stack_depth: 15
+  stem_kernel_size: [5, 4, 4]
```
mattersoflight, line 17: Shouldn't this section have an argument for checkpoint?

ziw-liu:
> I've turned on the merge commits option. git log and git blame should be more intuitive to read.
There is a trade-off with git-blame granularity. With squashed commits it's easier to link a change to a PR, which in the longer term has been more useful for me. Also people tend to write PRs better than they would write each of the commit messages, and it is difficult to enforce descriptive commit messages without manually rewriting history very often.

applications/contrastive_phenotyping/contrastive_cli/plot_embeddings.py
```diff
@@ -0,0 +1,109 @@
...
+    y=embedding[:, 1],
+    kind="hist",
+    col=infection,
+    bins=64,
+    cmap="inferno",
+)
+
+# %%
+# interactive scatter plot to associate clusters with specific cells
+scatter(
```
Soorya19Pradeep, line 92: I was able to find the cells in interactive plot using the track id of cells.
```
# interactive scatter plot to associate clusters with specific cells
scatter(
    x=embedding[:, 0],
    y=embedding[:, 1],
    color=(infection.astype(str) + " " + division.astype(str)),
    hover_name=features["fov_name"] + "/" + features["track_id"].astype(str) + "/" + features["id"].astype(str),
)
```
It will be good to have the time point detail as well, so that we can be sure.

ziw-liu: I added them as data fields.

Soorya19Pradeep: Thanks @ziw-liu! I have tested it. It works!

Soorya19Pradeep (approved):
I have tested the script and generated the umaps for exploring the data

applications/contrastive_phenotyping/contrastive_cli/predict.yml
```diff
@@ -0,0 +1,44 @@
...
+  accelerator: gpu
+  strategy: auto
+  devices: auto
+  num_nodes: 1
+  precision: 32-true
+  callbacks:
+    - class_path: viscy.light.embedding_writer.EmbeddingWriter
+      init_args:
+        output_path: 2024_02_04-tokenized-drop_path_0_0.zarr
+  inference_mode: true
```
mattersoflight, line 12: lightning_logs are written in current directory and not the parent of `output_path`. I suggest changing that behavior so that zarr stores with embeddings and lightning_logs are in the same folder.

mattersoflight:
Interactive prediction is quite useful. I also tested the prediction CLI. If feasible, please write `lighting_logs` during prediction to the parent directory of `output_path`.

applications/contrastive_phenotyping/contrastive_cli/plot_embeddings.py
```diff
@@ -0,0 +1,121 @@
...
+
+# %%
+# cluster features in heatmap directly
+inf_codes = pd.Series(infection.values.codes, name="infection")
+lut = dict(zip(inf_codes.unique(), "brw"))
+row_colors = inf_codes.map(lut)
+
+g = sns.clustermap(
+    scaled_features, row_colors=row_colors.to_numpy(), col_cluster=False, cbar_pos=None
+)
```
mattersoflight, line 119: I notice a warning: `Clustering large matrix with scipy. Installing `fastcluster` may give better performance.
`

applications/contrastive_phenotyping/contrastive_cli/plot_embeddings.py
```diff
@@ -0,0 +1,121 @@
...
+    kind="hist",
+    col=infection,
+    bins=64,
+    cmap="inferno",
+)
+
+# %%
+# interactive scatter plot to associate clusters with specific cells
+
+px.scatter(
```
mattersoflight, line 100: Use of plotly throws the following error in my environment: 
```
ValueError: Mime type rendering requires nbformat>=4.2.0 but it is not installed

```

But, I still get the plots

applications/contrastive_phenotyping/contrastive_cli/plot_embeddings.py
```diff
@@ -0,0 +1,144 @@
...
+        "y": "sample",
+        "color": "value",
+    },  # change labels to match our metadata
+    y=sample_id,
+    # show fov_name as y-axis
+)
+
+# %%
+scaled_features = StandardScaler().fit_transform(features.values)
+
```
mattersoflight, line 50: What is the rationale for this scaling? We are noticing in the raw embeddings that the cell division events have different mean/std. We should preserve that difference.

mattersoflight:
@ziw-liu I think the prediction and plotting are in good shape! You can squash and merge the current state.

I am summarizing outstanding comments, which should be handled in the next PR:
* Please write lighting_logs during prediction to the parent directory of output_path of predictions.
* Consider using fastcluster? I notice a warning: Clustering large matrix with scipy. Installing fastcluster may give better performance.
* Most importantly, does z-scoring of features collapse clusters?

applications/contrastive_phenotyping/contrastive_cli/predict.yml
ziw-liu: The log location is controlled separately by the logger callback, which assumes the default behavior here.

applications/contrastive_phenotyping/contrastive_cli/plot_embeddings.py
ziw-liu: I took this from the library's [example](https://umap-learn.readthedocs.io/en/latest/basic_usage.html). If I understand correctly this normalize per feature, not per sample. So for example a feature has 10x lower variance in all samples, it will bring it to the same importance as other features. So if a sample have different mean/std across all features, it should be preserved.

PR #143, opened 2024-08-23
Use new AMP custom forward API

ziw-liu:
Fix a deprecation warning.


mattersoflight:
@ziw-liu please fix this commit and expand on the intent of the PR

ziw-liu:
Closing in favor of #151 so that the fix is based on main.

PR #144, opened 2024-08-25
Add script to visualize effective receptive field

ziw-liu:
Add a script to visualize effective receptive field using point response in back propagation.

A vanilla U-Net after initialization:
image

UNeXt2 (2D) after initialization:
image

Trained CytoVS2D:
image


edyoshikun (approved):
LGTM! super useful

PR #141, opened 2024-08-17
SSL: fix MLP head and remove L2 normalization

mattersoflight:
Fix the sequence of batchnorm and linear in the last MLP layer of the contrastive encoder model.

Remove L2 normalization before computing triplet loss. This works best when also reducing the dimension of projections.

Refactor `light` module into `representation` and `translation` to separate pipelining code for different tasks.

Fix #139, fix #138.

Issue #123, opened 2024-08-05
Time sampling for positive pair 


alishbaimran:
**Goal**: be able to sample the positive pair given a "timestep". For example, sample at timestep 1 would mean sampling the nearest timepoint. 

Key features:
- [ ] Specify range of timestep to sample. Timestep 1 means nearest timepoint, timestep 2 means second nearest timepoint, etc.
- [ ] Use sampling rate (should be specific depending on dataset) to determine the experimental timepoint to draw anchor pair from.
- [ ] Sample either forward or backward. We can pick a direction for now and always sample that as this theoretically shouldn't impact the prediction. If we are at the edge and don't have timepoint before or after, then sample the other way. If timepoints available both forward and backward, then always sample x (forward or backward).

PR #145, opened 2024-08-28
SSL: fix MLP head and remove L2 normalization

### LINKED ISSUE #138: Hard-coded stem stride in `ContrastiveEncoder`
Currently the `stem_stride` parameter is not used and the stem stride will be the same (5, 4, 4) regardless of the kernel size. There is also no way to specify the stride from the lightning module (and the config).

This was due to a incomplete refactor in #130.

### LINKED ISSUE #139: Update the projection head (normalization and size).
TL;DR: current projection head doesn't do what it is supposed to do. It should have a batch norm and the size of the features and projections may be reduced further.
In the previous implementations of contrastive learning models (`dynacontrast`), we used batch norm in the projection head after each MLP.
```
(projection): Sequential(
    (fc1): Linear(in_features=2048, out_features=2048, bias=False)
    (bn1): BatchNorm1d(2048, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (relu1): ReLU()
    (fc2): Linear(in_features=2048, out_features=128, bias=False)
    (bn2): BatchNorm1dNoBias(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
  )
```

This [paper](https://arxiv.org/pdf/2212.11491) also recommends using a non-linear projection head with batch norm. Different projection heads are evaluated by comparing the rank of the features (# of independent features) before and after the projection head:
image

As expected, the rank(projections) << rank(features). 

[Our current model's behavior](https://github.com/mehta-lab/VisCy/tree/cbf280f445a3675a3b82a298cdf873935700b894) is the opposite: rank(projections)> rank(features) as seen from the examination of the principal components in each.

image

This seems to be the consequence of clipping of projections, which seems to be due to the use of ReLU without normalization.

`plt.plot(np.mean(embedding_dataset["projections"].values,axis=1))` 
image

`plt.plot(np.std(embedding_dataset["projections"].values,axis=1))`
image

ziw-liu:
Fix the sequence of batchnorm and linear in the last MLP layer of the contrastive encoder model.

Remove L2 normalization before computing triplet loss. This works best when also reducing the dimension of projections.

Refactor `light` module into `representation` and `translation` to separate pipelining code for different tasks.

Fix #139, fix #138.

Reopened from #141 to rename base branch.


mattersoflight (approved):
Thanks @ziw-liu for this refactor. Reorganization in `viscy.data`, `viscy.representation` and `viscy.translation` makes a lot of sense. Since contrastive models have been trained with this branch, makes sense to merge into `representation` branch. I am running a training, but haven't evaluated it.

TODO before merging:
@Soorya19Pradeep please test that you can train a model with this branch
@ziw-liu  we need following clean-up.
* Training and prediction scripts in applications.contrastive_phenotyping are obsolete and should be deleted.
* There are two versions of DLMBL exercise `examples/virtual_staining/dlmbl_exercise` and `examples/virtual_staining/img2img_translation`. One of them is old and should be deleted.

After merging, @ziw-liu, please test that refactor has not broken the image translation task. I ran a few scripts (`count_flops.py`, `network_diagram.py`), but we need proper testing with a few trainings.

Soorya19Pradeep:
@ziw-liu , @mattersoflight , I was able to set off training as well. It has continued for a few epochs for me as well.

mattersoflight:
@ziw-liu This is ready to merge into `representation`. After this merge, please continue building the time sampling strategy on the `representation` branch. Let's discuss the design in #123.

PR #154, opened 2024-08-31
Sample positive and negative samples with a time offset for the triplet contrastive task

### LINKED ISSUE #123: Time sampling for positive pair

ziw-liu:
Sampling strategy for triplet contrastive learning is now configured with `viscy.data.triplet.TripletDataModule(time_interval=...)`.

1. `time_interval == "any"`
Same as before this PR. The positive sample is the augmented version of the anchor, and the negative sample is any other patch from a different track.
If each node (`id`) in the tracking result is assigned a different `track_id`, this reduces to the classical patch-wise contrastive learning without tracking information.
2. `isinstance(time_interval, int) == True`
For time interval $\tau$, the positive sample of anchor at time point $t$ is sampled from the same track at time point $t + \tau$, and the negative example is sampled from a different track also at time point $t + \tau$. Example log from CTC's DIC-D2DH-HeLa dataset (anchor, positive, negative):

image

Closes #123.

### DIFF
--- a/applications/contrastive_phenotyping/contrastive_cli/fit_ctc_mps.yml
+++ b/applications/contrastive_phenotyping/contrastive_cli/fit_ctc_mps.yml
@@ -1,0 +1,113 @@
+# See help here on how to configure hyper-parameters with config files:
+# https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_advanced.html
+seed_everything: 42
+trainer:
+  accelerator: gpu
+  strategy: auto
+  devices: 1
+  num_nodes: 1
+  precision: 32-true
+  logger:
+    class_path: lightning.pytorch.loggers.TensorBoardLogger
+    # Nesting the logger config like this is equivalent to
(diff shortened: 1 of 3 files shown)


mattersoflight (approved):
LGTM! I read through the code, but haven't had a chance to test it.

PR #159, opened 2024-09-10
Fixes on correlation of PCA and UMAP components to computed_feature script

Soorya19Pradeep:
I am adding small bug fixes on the script for computing the correlation between PCA and UMAP features and the computed features on this branch.


ziw-liu:
@Soorya19Pradeep do you expect more work on this branch?

Soorya19Pradeep:
> @Soorya19Pradeep do you expect more work on this branch?

Yes @ziw-liu . I am going through, correcting the feature extraction methods

Soorya19Pradeep:
@ziw-liu, what is the best way to combine the `PC_vs_CF.py` and `PC_vs_CF_singleChannel.py` scripts? One uses model trained on both phase and sensor channel and the other use just phase channel.

ziw-liu:
> what is the best way to combine the `PC_vs_CF.py` and `PC_vs_CF_singleChannel.py` scripts? One uses model trained on both phase and sensor channel and the other use just phase channel.

You can put the fluorescence features behind a condition that only executes when there is a second channel.

ziw-liu:
@Soorya19Pradeep we decided to not use UMAP for feature correlation. Can you update this PR to reflect that?

PR #160, opened 2024-09-12
Linear probing

ziw-liu:
~Refactored the linear classifier with lightning.~
Use logistic regression to optimize a linear classifier and copy the weights to torch linear layer (#169).
Fixed the bug where the dataset is not split: the dataset is now split manually by FOVs so that we can count infection percentage only in the test FOV.
Added a script to demonstrate usage.
Added a script to use the trained classifier for visualizing integrated gradients.


alishbaimran:
Haven't tested code but looked through the script which looks good.

Soorya19Pradeep:
@ziw-liu , I have tested the code. It works well for me. I was able to reproduce similar numbers as what I saw with my earlier iteration. Thanks!

PR #161, opened 2024-09-12
Adding a caching dataloader for 3D dataset

edyoshikun:
## Summary
This PR adds a dataloader that caches the datasets in RAM and avoids using the `SliceWindowDataset()`. Here we cache the dataset on epoch 1. Ideally, the datasets fit within the 2TBs available per node. One can use `btop` to inspect the resource usage.

The z-slicing for the volumes is set as a parameter `num_z_slices`. We leave the slicing to the configurable MONAI transforms. 

### Working 
- [x] Tested the `hcs_ram.CachedDataset`
- [x] Tested `hcs_ram.CacheDataModule`
- [x] Tested running `UNeXt2` with a small toy dataset locally (M1) and `gpu-sm02` nodes.
- [x] Running the model with `dev_mode=True`, which runs the model for 1 epoch.

### Not Working 
- Running UNeXt2 with the actual datasets (500GB+) using one GPU (H100), one node
- Running UNeXt2 with the actual datasets (500GB+) using 4 GPU (H100), one node

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ license = { file = "LICENSE" }
 authors = [{ name = "CZ Biohub SF", email = "compmicro@czbiohub.org" }]
 dependencies = [
     "iohub==0.1.0",
-    "torch>=2.1.2",
+    "torch>=2.4.1",
     "timm>=0.9.5",
     "tensorboard>=2.13.0",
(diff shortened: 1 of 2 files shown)


ziw-liu:
Closing in favor of #195.

PR #163, opened 2024-09-16
smote and save predictions

alishbaimran:
- Replaced torch.amp.custom_fwd with torch.cuda.amp.custom_fwd in engine.py for translation.
- Fixed linear classifier to include SMOTE for training and ability to save predictions for figure use.
- Complete example script that can be used for training classifier and generating % infected cells figure.


viscy/representation/lca.py
```diff
@@ -68,27 +71,50 @@ def __init__(
...
+        train_dataset, val_dataset, test_dataset = torch.utils.data.random_split(
+            self.dataset, [train_size, val_size, test_size]
         )
 
+        self.test_indices = [self.dataset[i][2].item() for i in test_dataset.indices]
+        self.train_dataset = [(x, y) for (x, y, _) in train_dataset]
+        self.val_dataset = [(x, y) for (x, y, _) in val_dataset]
+        self.test_dataset = [(x, y) for (x, y, _) in test_dataset]
+
+        if self.use_smote and stage == "fit":
```
ziw-liu, line 103: If `use_smote == False` this will be a silent error.

viscy/translation/engine.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         self.l2_alpha = l2_alpha
         self.ms_dssim_alpha = ms_dssim_alpha
 
-    @torch.amp.custom_fwd(device_type="cuda", cast_inputs=torch.float32)
+    @torch.cuda.amp.custom_fwd(cast_inputs=torch.float32)
```
ziw-liu, line 71: Merge error.

viscy/representation/lca.py
```diff
@@ -188,6 +221,13 @@ def train_and_test_linear_classifier(
         Batch size, by default 1024
     lr : float, optional
         Learning rate, by default 1e-3
+    use_smote : bool, optional
+        Whether to apply SMOTE to the training data, by default False
+    save_predictions : bool, optional
+        Whether to save predictions to CSV, by default False
+    csv_path : str, optional
+        Path to save predictions CSV, by default None
+    merged_df : pd.DataFrame, optional
```
ziw-liu, line 230: What should this be?

viscy/representation/lca.py
```diff
@@ -197,8 +237,33 @@ def train_and_test_linear_classifier(
...
+        batch_size,
+        use_smote=use_smote,
     )
     model = LinearClassifier(embeddings.shape[1], num_classes, lr)
     trainer.fit(model, data)
     trainer.test(model, data)
+
+    test_preds = model.test_predictions_cache
+
+    if save_predictions and csv_path is not None and merged_df is not None:
```
ziw-liu, line 252: Silent error if not all three are true.

viscy/representation/lca.py
```diff
@@ -68,27 +71,50 @@ def __init__(
...
+            Whether to apply SMOTE to the training data, by default False
         """
         super().__init__()
         if not embeddings.shape[0] == labels.shape[0]:
             raise ValueError("Number of samples in embeddings and labels must match.")
         if sum(split_ratio) != 1.0:
             raise ValueError("Split ratio must sum to 1.")
-        self.dataset = TensorDataset(embeddings.float(), labels.long())
+        indices = torch.arange(embeddings.shape[0])
+        self.dataset = TensorDataset(embeddings.float(), labels.long(), indices)
```
ziw-liu, line 83: This is not used? `torch.utils.data.Subset` has `indices` as an attribute.

viscy/representation/lca.py
```diff
@@ -68,27 +71,50 @@ def __init__(
...
+        self.train_dataset = [(x, y) for (x, y, _) in train_dataset]
+        self.val_dataset = [(x, y) for (x, y, _) in val_dataset]
+        self.test_dataset = [(x, y) for (x, y, _) in test_dataset]
+
+        if self.use_smote and stage == "fit":
+            train_embeddings, train_labels = zip(*self.train_dataset)
+            train_embeddings = torch.stack(train_embeddings)
+            train_labels = torch.tensor(train_labels)
+
+            smote = SMOTE(random_state=42)
```
ziw-liu, line 108: Seed should be set in the user script.

ziw-liu: i.e. `lightning.pytorch.seed_everything(42, workers=True)`

viscy/translation/engine.py
ziw-liu: I'm having a hard time understanding this. #151 is merged and released with v0.2.1. Why are you changing this in this PR?

alishbaimran: Test won't pass if I don't change this. I can revert the change but then you'd have to be aware that some of the tests won't pass on this PR.

viscy/representation/lca.py
alishbaimran: csv_path is where to same predictions and merged_df is the dataframe of initial input data. You can see the example script to see functionality of this.

viscy/representation/lca.py
alishbaimran: Hm ok, I need to know the indicies of the test set from the original dataset such that I can know which cells to populate predictions for in the output csv. Is there another approach to do this?

I am using indicies in train_and_test_linear_classifier (line 253)

test_indices = data.test_indices

viscy/translation/engine.py
ziw-liu: This is because your branch is not up to date with the base branch. `train-linear-classifier` already have this:
https://github.com/mehta-lab/VisCy/blob/4fe26108cf14bc17a4b3c8b3c484eb6ef26f1307/viscy/translation/engine.py#L71

viscy/representation/lca.py
ziw-liu: Can you include the above information in the docstring?

viscy/representation/lca.py
ziw-liu: That is exactly what I'm referring to. You are using the attribute of `Subset`, which is not the value of the dataset. The third tensor value is never used.

ziw-liu: One idea is to do the class balancing and splitting outside in numpy/sklearn. That way we can re-use this for the sklearn models (logistic regression, KNN etc) too.

alishbaimran: Do you want to refactor to include this?

alishbaimran: I was going to do this but then you'd have to change LCA.py more as the splitting would be done manually outside in the script. If you could refactor to include this and make sure functionality of being able to save a csv stays the same, that would be good.

ziw-liu: Do you still want to fix this PR and merge it before refactoring?

alishbaimran: Yes, that would be ideal so we can start making figures with it. What is the minimal change here for the indicies you'd like me to do?

ziw-liu: Use the `Subset.indices` attribute directly and avoid adding the indices as a value of the dataset.

PR #169, opened 2024-09-19
Use sklearn's logistic regression for linear probing

ziw-liu:
Optimizing the logistic regression model in sklearn (max-entropy) is the exact same function and objective as optimizing a linear layer with cross-entropy-with-logit loss in torch. The only differences are least-squares vs SGD, and weight regularization. Therefore, by copying the weights and bias from a sklearn model to a torch linear layer, we can use the same deterministic optimization for evaluating classification performance and model interpretation methods that require gradients.

I also include a test case to verify the above: the logits produced by the sklearn and torch models are almost equal (with relative tolerance 0.001).

### DIFF
--- a/applications/contrastive_phenotyping/evaluation/grad_attr.py
+++ b/applications/contrastive_phenotyping/evaluation/grad_attr.py
@@ -10,13 +10,15 @@ from cmap import Colormap
 from lightning.pytorch import seed_everything
 from skimage.exposure import rescale_intensity
 from sklearn.linear_model import LogisticRegression
-from sklearn.preprocessing import StandardScaler
 
 from viscy.data.triplet import TripletDataModule
 from viscy.representation.embedding_writer import read_embedding_dataset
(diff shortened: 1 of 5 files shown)


ziw-liu:
Example output of `applications/contrastive_phenotyping/evaluation/linear_probing.py`:

```
Trained logistic regression classifier.
Training set accuracy:
              precision    recall  f1-score   support

           0      0.997     0.999     0.998      3836
           1      0.998     0.994     0.996      2076

    accuracy                          0.997      5912
   macro avg      0.997     0.997     0.997      5912
weighted avg      0.997     0.997     0.997      5912

Test set accuracy:
              precision    recall  f1-score   support

           0      0.963     0.987     0.975      4251
           1      0.977     0.938     0.957      2551

    accuracy                          0.968      6802
   macro avg      0.970     0.962     0.966      6802
weighted avg      0.968     0.968     0.968      6802
```

PR #170, opened 2024-09-20
Tweak attribution visualization

ziw-liu:
Tweaked plotting of attribution.

@Soorya19Pradeep @alishbaimran I added a matplotlib style sheet that will produce the exact typeface and fontsize in latex for your reference. For a script in `applications/contrastive_phenotyping/evaluation`, it can be used with:

```py
plt.style.use("./figure.mplstyle")
```

### DIFF
--- a/applications/contrastive_phenotyping/evaluation/figure.mplstyle
+++ b/applications/contrastive_phenotyping/evaluation/figure.mplstyle
@@ -1,0 +1,9 @@
+font.family: sans-serif
+font.sans-serif: Arial
+font.size: 10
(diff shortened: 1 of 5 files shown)


ziw-liu:
@Soorya19Pradeep @alishbaimran UMAP scatter plot like this can achieve a density-like effect, good for overlaying tracks on:

https://github.com/mehta-lab/VisCy/blob/e423b2a932c6c582a4400af2e81bf44d5c61a512/applications/contrastive_phenotyping/figures/organelle_dynamics.py#L161-L163

PR #171, opened 2024-09-22
Test banner tag in readme

ziw-liu:
Test an idea, don't need to merge:
Banner tags like `> [!NOTE]` render nicely on github, but probably won't render in other places since it's not CommonMark. See the README on this branch (view file).

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,8 @@ This repository provides a pipeline for the following.
 - Image representation learning
   - Self-supervised learning of the cell state and organelle phenotypes
 
-> **Note:**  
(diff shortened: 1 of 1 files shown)


edyoshikun (commented):
I don't think this will render in pypi. I was reading and this syntax is specific to github,sphinx,etc, as you say and won't work for pypi's markdown style (see image below to see the rendering).

I prefer the current way, but perhaps add some emoji ⚠️ℹ️ (?)

Screenshot 2024-09-23 at 8 58 50 PM

PR #181, opened 2024-10-10
CLI interface for multiple learning tasks

### LINKED ISSUE #127: CLI for multiple training tasks
As part of #126, @ziw-liu suggested:  
> For the future iteration (v0.3?) it might make sense to replace the manual CLI entry point (viscy [command]) with python's built-in (python -m viscy.cli.module [command]) to make supporting multiple learning tasks easier.

This makes a lot of sense and I'd vote to make this change sooner to clarify that the repo does implement multiple learning tasks. It is easy to change our virtual training scripts that launch training from `viscy [command]` -> `python -m viscy.cli.module [command]`. Are there other factors for which we should wait to refactor CLI? 

@ziw-liu @edyoshikun

ziw-liu:
Close #127.

## State of this PR

A different CLI is used per task. The global `viscy [command]` entry point is removed, and tasks can be invoked by:

```sh
# preprocessing is the same for both tasks
python -m viscy.translation preprocess --data_path /path/to/data.zarr --num_workers=8
python -m viscy.representation preprocess -c /path/to/config.yaml

# different tasks will use hard-coded trainer, model, and data classes
python -m viscy.translation fit -c /path/to/config.yaml
python -m viscy.representation predict -c /path/to/config.yaml
```

## TODO and discussion

- Update docs and shell scripts.

- Use [dependency injection](https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_advanced_3.html#multiple-models-and-or-datasets) for data module and model classes. For example, this allows virtual staining training from CLI for not just the currently hard-coded `HCSDatamodule`, but also the combined/concatenated datasets, and other implementations such as #161. Another use case would be to run different SSL tasks than triplet (e.g. SimCLR, BYOL) without adding a CLI entry point each time. See #182 for examples.

After the dependency injection for data and model classes is implemented, it is then possible to keep the global `viscy [command]` CLI and dispatch to tasks through configs (see draft implementation in #182). Currently this will work except the minor caveat of not being able to use different trainers per task, resulting in [2 redundant lines](https://github.com/mehta-lab/VisCy/blob/76b66be9519785f5c38d8931186ede86174e79a7/viscy/translation/__main__.py#L15-L16) in VS configs, but _might_ pose more problems in the future if there needs to be more significant divergence. @mattersoflight @edyoshikun Which one (`python -m viscy.[task] [command]` or `viscy [command]`) do you prefer?


mattersoflight:
@ziw-liu Dependency injection via configs and simplifying the CLI structure to `viscy [command]` makes the most sense.  We can fully document and hand over complex training and prediction recipes via configs. Let's go this route.

ziw-liu:
> Are you saying that if we use configs to specify the task and the command (step in the task), we cannot configure the [trainer](https://lightning.ai/docs/pytorch/stable/common/trainer.html) with callbacks via config?

We can still configure all the parameters as before. Different learning tasks will have to share the same `Trainer` and `LightningCLI` classes. When using different CLI entry points, it's possible to use different CLI subclasses ([1](https://github.com/mehta-lab/VisCy/blob/76b66be9519785f5c38d8931186ede86174e79a7/viscy/representation/__main__.py#L11), [2](https://github.com/mehta-lab/VisCy/blob/76b66be9519785f5c38d8931186ede86174e79a7/viscy/translation/__main__.py#L21)). This was used to automatically link task-specific CLI arguments that are shared between the model and data groups. Without this, there will be some redundant lines in the config (currently 2 for virtual staining).

ziw-liu:
Closing in favor of #182.

PR #196, opened 2024-10-23
GPU transform for FCMAE pre-training

ziw-liu:
Building on top of #196:
Doing non-trivial augmentations on large imaging volumes has become a bottleneck in training image translation models. By keeping the initial cropping and normalization on the CPU workers, while executing the more heavy transforms (especially the resampling ones) on the GPU, training can be significantly faster.

## Current state of this PR

For 2D FCMAE, reaching the same validation loss is ~7x the speed compared to v0.2. Compare these logs:

```
/hpc/projects/comp.micro/virtual_staining/models/hek-a549-bj5a-20x/lightning_logs/tiny-2x2-fcmae-amp-hek-a549-bj5a-400ep
/hpc/projects/comp.micro/virtual_staining/models/fcmae-2d/test-logs/hek-a549-bj5a-gpu-aug-val-every-10
```

And for 3D FCMAE, the total voxel throughput is ~3x of v0.2, potentially limited by CPU-GPU transfer.

~Caveat: since the transforms are now defined in the lightning module instead of the data module, they are the same for all the dataset.~ (fixed)

### DIFF
--- a/viscy/data/combined.py
+++ b/viscy/data/combined.py
@@ -28,7 +28,7 @@ class CombinedDataModule(LightningDataModule):
     train_mode : CombineMode, optional
         mode in training stage, by default CombineMode.MAX_SIZE_CYCLE
     val_mode : CombineMode, optional
-        _description_, by default CombineMode.SEQUENTIAL
+        mode in validation stage, by default CombineMode.SEQUENTIAL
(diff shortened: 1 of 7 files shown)


edyoshikun (approved):
mrnging this so we can merge to base

PR #202, opened 2024-11-07
Fix validation loss aggregation in VSUNet

ziw-liu:
Fix a bug where the training loop does not clear `loss/validate` metrics after logging it. The logged metric will be the average of all the epoch, affecting checkpoint selection.

### DIFF
--- a/viscy/translation/engine.py
+++ b/viscy/translation/engine.py
@@ -378,7 +378,6 @@ class VSUNet(LightningModule):
     def on_validation_epoch_end(self):
         super().on_validation_epoch_end()
         self._log_samples("val_samples", self.validation_step_outputs)
-        self.validation_step_outputs = []
         # average within each dataloader
         loss_means = [torch.tensor(losses).mean() for losses in self.validation_losses]
         self.log(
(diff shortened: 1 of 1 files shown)


ziw-liu:
This bug means that previously, as long as the loss of the new $N$-th epoch $l_N$ is lower than the average loss than the previous epochs, the new 'validation loss' logged (which is actually the average over $N$ epochs, $\overline{l}_{N}$), will appear to decrease:

```math
\overline{l}_{N} - \overline{l}_{N-1}
= \frac{\sum_{i=0}^N l_i}{N} - \frac{\sum_{i=0}^{N-1} l_i}{N-1}
= \frac{1}{N} (l_N - \frac{1}{N-1} \sum_{i=0}^{N-1} l_i)
= \frac{1}{N} (l_N - \overline{l}_{N-1})
```

ziw-liu:
These being said, we also watch per-step validation losses. They are a bit noisy, but I don't think our models have major overfitting issues.

edyoshikun (approved):
more users more bugs! thanks for the fix.

PR #211, opened 2024-11-21
Rank nearest neighbors in embedding space

ziw-liu:
An ideal cell state representation has the following properties:

- When the cell state does not change, the neighboring time points of the same cell should have more similar embeddings compared to other cells (continuity)
- When the cell state does change, the embeddings change a lot so that different cell states can be expressed (dynamic range)

We observed these properties for time-regularized models visually in UMAPs. However, Euclidean distances in UMAP space carry no global meaning, so the observations cannot be directly quantified by comparing UMAP values. This PR provides methods to describe the embedding similarities close to how UMAP is computed: each sample is ranked as the $k$-th nearest neighbor for each sample, and the displacement in this neighborhood ($k$ at $t_{i+1}$ for each $t_i$) can then be used to describe the fluctuation of embeddings in a way that preserves latent space topology.

Example from ALFI dataset:

image

### DIFF
--- a/applications/contrastive_phenotyping/evaluation/time_decay_knn.py
+++ b/applications/contrastive_phenotyping/evaluation/time_decay_knn.py
@@ -1,0 +1,95 @@
+# %%
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+from viscy.representation.embedding_writer import read_embedding_dataset
+from viscy.representation.evaluation.clustering import (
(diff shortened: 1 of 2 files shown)


ziw-liu:
Using standardized features changes the angular distance, but doesn't really change the shape of the rankings curve:

image

edyoshikun (approved):
Works like a charm! Also super helpful and thanks for this elegant implementation.

PR #182, opened 2024-10-10
Restore global CLI entry point

ziw-liu:
Use the global `viscy [command]` CLI for multiple tasks (translation/representation/...) and dispatch via dependency injection in the config.

Example config for VSCyto2D prediction:

```
/hpc/projects/intracellular_dashboard/ops/2024_09_19_tracking_accuracy_test/2-VS/tta/test_cli/predict.yml
```

PR #212, opened 2024-11-21
Explicit target shape argument in the HCS data module

ziw-liu:
Replace the `architecture` argument with explicit `targe_2d` to avoid confusion. Also changed the default to `False` since 2.5D is no longer used in production.

Previously, if `architecture` is not specified, it will default to '2.5D' and produce incorrect results for 3D models, where the output would have 1 Z-window worth of extra blank slices relative to the input. This was not a problem before #182, when this argument was hard-linked by the CLI with the model configuration.

@tayllatheodoro @Soorya19Pradeep @ieivanov @talonchandler:
This parameter used to be automatically configured (correctly) for `viscy<=0.2.1`, which is all the stable versions on PyPI. If you recently did 3D predictions with a pre-release installation, please check the results to make sure that the configuration is correct (look for blank slices at the start of z-stacks). Specifically for @tayllatheodoro I noticed that some of the mantis predictions was affected by this.

### DIFF
--- a/tests/data/test_data.py
+++ b/tests/data/test_data.py
@@ -58,7 +58,7 @@ def test_datamodule_setup_fit(preprocessed_hcs_dataset, multi_sample_augmentatio
         batch_size=batch_size,
         num_workers=0,
         augmentations=transforms,
-        architecture="3D",
+        target_2d=False,
         split_ratio=split_ratio,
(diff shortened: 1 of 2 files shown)


edyoshikun (approved):
LGTM! ship it 🚀

PR #216, opened 2025-01-02
Fixing the TripletDataModule() after changes to HCSDataModule

edyoshikun:
We forgot to change the TripletDataModule() after the HCSDatamodule() merged changes in #212

### DIFF
--- a/viscy/data/triplet.py
+++ b/viscy/data/triplet.py
@@ -339,7 +339,7 @@ class TripletDataModule(HCSDataModule):
             split_ratio=split_ratio,
             batch_size=batch_size,
             num_workers=num_workers,
-            target_2d=bool(z_range[1] - z_range[0] == 1),
+            target_2d=False,
             yx_patch_size=final_yx_patch_size,
             normalizations=normalizations,
             augmentations=augmentations,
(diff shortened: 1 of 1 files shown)


viscy/data/triplet.py
```diff
@@ -339,7 +339,7 @@ def __init__(
             split_ratio=split_ratio,
             batch_size=batch_size,
             num_workers=num_workers,
-            architecture="UNeXt2",
+            target_2d=bool(z_range[1] - z_range[0] == 1),
```
ziw-liu, line 342: Shouldn't this be always false?

edyoshikun:
Not sure if we should also fix these in the files under `/applications/infection_classification`. Are these still used @Soorya19Pradeep ?

https://github.com/mehta-lab/VisCy/blob/bc3621755209107dd3dac3f3a29e7182576ba803/applications/infection_classification/Infection_classification_covnextModel.py#L32

viscy/data/triplet.py
ziw-liu: Since there is no target anyway.

Soorya19Pradeep (approved):

PR #195, opened 2024-10-21
Sharded distributed sampler for cached dataloading in DDP

ziw-liu:
Add a distributed sampler that only permutes index within ranks, improving cache hit rate in DDP.

See `viscy/scripts/shared_dict.py` for usage.

Also includes changes from #196.

PR #218, opened 2025-01-06
Memory-mapped caching for image translation training

ziw-liu:
#195 implemented in-RAM caching for image translation trainings. However it does not scale to datasets larger than available system memory. This PR implements a node-local disk cache via [tensordict's memory-mapped tensor](https://pytorch.org/tensordict/stable/reference/generated/tensordict.MemoryMappedTensor.html#tensordict.MemoryMappedTensor).


ziw-liu:
Progress summary: previously training on 2.3 TB of datasets takes 20 s/iter, now training on 3 TB takes 10 s/iter.

Lessons learned:
- Default local scratch configuration on our compute nodes was not optimal for sustained I/O (ZFS and very small sector size). This was fixed on select nodes and will be rolled out more broadly later.
- Optimizations/mitigations done for system memory caching might hurt performance in the mmap setup. Moving some transforms back to the CPU improved end-to-end timing. This might be because MONAI transforms are not batched (executed in a loop), and CPU/GPU sync could be taking much longer than the actual compute.

To be investigated:
- Moving augmentations back to the CPU recreated the CPU compute bottleneck. Potentially fixable by using batched augmentation or distributing the compute better across devices.
- Precomputing the normalization to simplify training-time logic (https://github.com/mehta-lab/VisCy/pull/218/commits/d2cd3401590c2204e24e98a270fa17a6d153d5d3).

ziw-liu:
> This might be because MONAI transforms are not batched (executed in a loop), and CPU/GPU sync could be taking much longer than the actual compute.

Benchmark in https://github.com/mehta-lab/VisCy/pull/218/commits/2e12f0022e65eceefe0d86fd0e32382097f37366 (10 runs, milliseconds):

| device | MONAI (sequential) | Kornia (batched) | Relative |
| :---: | :--: | :---: | :---: |
| 16 Zen 2 CPUs | 7320 | 556 | 13.2 |
| A40 GPU | 2620 | 210 | 12.5 |

edyoshikun:
Marking this ready for review. Taking a look now! @ziw-liu

viscy/data/mmap_cache.py
```diff
@@ -0,0 +1,255 @@
...
+    train_cpu_transforms : list[DictTransform]
+        Transforms to be applied on the CPU during training.
+    val_cpu_transforms : list[DictTransform]
+        Transforms to be applied on the CPU during validation.
+    train_gpu_transforms : list[DictTransform]
+        Transforms to be applied on the GPU during training.
+    val_gpu_transforms : list[DictTransform]
+        Transforms to be applied on the GPU during validation.
+    pin_memory : bool, optional
+        Use page-locked memory in data-loaders, by default True
```
edyoshikun, line 126: missing prefetch_factor docstring

edyoshikun: same goes wiht the scratch, include_wells and exclude_fovs

viscy/data/gpu_aug.py
```diff
@@ -169,7 +173,23 @@ def __getitem__(self, idx: int) -> dict[str, Tensor]:
         return sample
 
 
-class CachedOmeZarrDataModule(GPUTransformDataModule):
+class SelectWell:
```
edyoshikun, line 176: what are your thoughts on setting a cache_limit? I guess we can assume we can assume we have all the resources possible and we can customize this with your include of FOV and exclusion of fovs. 
```python
def __getitem__(self, idx: int) -> dict[str, Tensor]:
    position, time_idx, norm_meta = self._metadata_map[idx]
    cache = self._cache_map[idx]
    if cache is None:
        _logger.debug(f"Loading volume for index {idx}")
        volume = torch.from_numpy(
            position[self.array_key]
            .oindex[time_idx, list(self.channels.values())]
            .astype(np.float32)
        )
        if not self.skip_cache:
            _logger.debug(f"Caching for index {idx}")
            self._cache_map[idx] = volume
            
            if self.max_cache_size is not None:
                self._update_cache_usage(idx)
                self._enforce_cache_limit()
    else:
        _logger.debug(f"Using cached volume for index {idx}")
        volume = cache
        if self.max_cache_size is not None:
            self._update_cache_usage(idx)
            
    sample = {name: img[None] for name, img in zip(self.channels.keys(), volume)}
    if self.load_normalization_metadata:
        sample["norm_meta"] = norm_meta
    if self.transform:
        sample = self.transform(sample)
    if not isinstance(sample, list):
        sample = [sample]
    return sample
def _update_cache_usage(self, idx: int) -> None:
    if idx in self._cache_usage_queue:
        self._cache_usage_queue.remove(idx)
    self._cache_usage_queue.append(idx)

def _enforce_cache_limit(self) -> None:
    if self.max_cache_size is None:
        return
        
    while len(self._cache_usage_queue) > self.max_cache_size:
        oldest_idx = self._cache_usage_queue.pop(0)
        if oldest_idx in self._cache_map and self._cache_map[oldest_idx] is not None:
            _logger.debug(f"Removing index {oldest_idx} from cache due to size limit")
            self._cache_map[oldest_idx] = None
```

viscy/preprocessing/precompute.py
```diff
@@ -0,0 +1,76 @@
...
+
+def _filter_wells(
+    plate: Plate, include_wells: list[str] | None
+) -> Generator[Well, None, None]:
+    for well_name, well in plate.wells():
+        if include_wells is None or well_name in include_wells:
+            yield well
+
+
+def _filter_fovs(
```
edyoshikun, line 37: here we cannot configure the FOVs per well. what this filters all fovs with the same name for all positions right?

edyoshikun: I am envisioning cases in like Mantis where fov `A/1/0000000` is good for one well but not for `A/2/000000` and we still want to use this fov.

viscy/preprocessing/precompute.py
```diff
@@ -0,0 +1,76 @@
...
+
+from pathlib import Path
+from typing import Generator, Literal
+
+import dask.array as da
+from dask.diagnostics import ProgressBar
+from iohub.ngff.nodes import Plate, Position, Well, open_ome_zarr
+
+
+def _normalize_image(
```
edyoshikun, line 11: this is so that we can compute the normalization per timepoint right?

ziw-liu: In principle yes, but it's currently not used in training.

viscy/preprocessing/precompute.py
ziw-liu: This is controlled by `include_wells` in the calling function below.

viscy/data/gpu_aug.py
edyoshikun: probably not, these are progressively loaded

edyoshikun (approved):
LGTM!

PR #219, opened 2025-01-08
Refactoring MSD, smoothness and dynamic range calculations and unecessary script cleanup

edyoshikun:
This PR refactors and adds the calculations for:

- [x] Mean square displacement
- [x] Slope calculation for smoothness
- [x] ~~Normalization strategies for the embeddings prior analysis~~ Decided to use raw
- [x] Cleaning up unecessary/obsolete scripts


viscy/representation/evaluation/distance.py
```diff
@@ -20,67 +21,184 @@ def calculate_cosine_similarity_cell(embedding_dataset, fov_name, track_id):
...
+        for tau, displacements in displacement_per_tau.items()
+    }
+    std_displacement_per_tau = {
+        tau: np.std(displacements)
+        for tau, displacements in displacement_per_tau.items()
+    }
+    return mean_displacement_per_tau, std_displacement_per_tau
 
 
 def compute_dynamic_range(mean_displacement_per_tau):
```
Soorya19Pradeep, line 204: @edyoshikun , does this way of computing dynamic range make sense? You talked about the effect of outliers

Soorya19Pradeep:
@edyoshikun , is there a way to switch off the display of the entire track and just display one time point where we click on the PC viewer app?

tests/representation/test_cellfeatures.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Statistical features should be positive but small
+        positive_features = ['mean_intensity', 'masked_intensity', 'std_dev', 'kurtosis', 'skewness', 'iqr', 'spectral_entropy', 
+                           'dissimilarity', 'contrast', 'texture', 'zernike_std', 'zernike_mean',
+                           'radial_intensity_gradient', 'intensity_localization', 'masked_area']
+        integer_features = ['masked_area']
+        float_features = list(set(positive_features) - set(integer_features))
+
+        for feature in positive_features:
+            assert abs(cell_features[feature].values[0]) >= 0, f"{feature} should be positive"
```
edyoshikun, line 77: this one as well. change since `abs()` returns positive values.

tests/representation/test_cellfeatures.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        return np.random.randint(0, 2, size)
+    
+    def test_constant_image_features_are_zero(self):
+        """Test that statistical features of a constant image are zero/constant"""
+        constant_image = self.create_constant_image()
+        binary_image = self.create_binary_image()
+
+        features = CellFeatures(constant_image, binary_image)
+        cell_features = features.compute_all_features()
```
edyoshikun, line 37: Since we have a nice strucure in the code where you can do `CellFeatures().compute_intensity_features()` I suggest we split and do one test function for each case. That way the tests are more focused, easier to debug and we can isolate the issues. In principle if we call the individual `compute_***()` we should be able to call `compute_all()` since this calls all of these in sequence. 


```python
CF = CellFeatures(image,mask)
features = CF.compute_intensity_features()

assert features["mean_intensity"] == value
assert features["std_dev"] == 0
assert features["min_intensity"] == value
assert features["max_intensity"] == value
```

I haven't written many tests myself with pytests, but we can also use either `hypothesis` or `pytest.mark.parametrize`. The latter I've used to make arrays with known input variables so we can re-use the functions like this
```python
        @pytest.mark.parametrize("value", [0, 1, 255, -100, 1e5])
        def test_constant_image_intensity_features(self, value, binary_mask):
            """Test intensity features with constant value images"""
            image = np.full((100, 100), value, dtype=np.float32)
            cf = CellFeatures(image, binary_mask)
            
            features = cf._compute_intensity_features()
            
            assert features["mean_intensity"] == value
            assert features["std_dev"] == 0
            assert features["min_intensity"] == value
            assert features["max_intensity"] == value
            assert features["kurtosis"] == 0  # Constant distribution has no kurtosis
            assert features["skewness"] == 0  # Constant distribution has no skewness
```

edyoshikun (commented):
@Soorya19Pradeep this is a good start. We should focus the tests to ensure that the functions handle a known input and behave as expected. By this I mean, if we throw in a known array with known features like you did using `random.uniform()` or similar, when we run individual functions in the class like `CellFeatures().compute_intensity_features()` do we get the expected mean,std,min,max,kurtosis,etc? We can use `pytest.approx(to the expected value)`. Most of these test, if you look at the `tests/data/test_data.py` are checking that we return the right shapes and numbers because that is crucial for feeding them into the model. In this case, we are more interested in the functionality of the function and its behavior.


I left individual comments, and happy to chat about each individual one.

viscy/representation/evaluation/visualization.py
```diff
@@ -0,0 +1,1733 @@
+# %%
```
ziw-liu, line 1: Is this file meant to be a script?

ziw-liu:
@edyoshikun is this still intended to be merged?

edyoshikun:
No, this branch is going to be closed in favor of #240

Issue #133, opened 2024-08-13
Move to NumPy style docstrings?


ziw-liu:
This package inherited single-line sphinx style docstrings from microDL. However this is different from the NumPy style that most of our contributors are used to and sometimes IDE settings get mixed up. Maybe moving to NumPy style will reduce some friction?

mattersoflight:
Good suggestion - I think we should switch over to NumPy style docstrings for the sake of readability of docstrings.

PR #220, opened 2025-01-29
Adding TArrow for testing

edyoshikun:
This PR adds the Tarrow package wrapped with lightning. It also contains a DataLoader to get the pair of timepoints from ome-zarr.

### DIFF
--- a/viscy/data/tarrow.py
+++ b/viscy/data/tarrow.py
@@ -1,0 +1,176 @@
+from pathlib import Path
+import numpy as np
+from iohub.ngff import Position, open_ome_zarr
+from lightning.pytorch import LightningDataModule
+from tarrow.data.tarrow_dataset import TarrowDataset
+from torch.utils.data import DataLoader, ConcatDataset
+import torch
+
+
+class TarrowDataModule(LightningDataModule):
+    def __init__(
+        self,
+        ome_zarr_path: str | Path,
+        channel_name: str,
+        train_split: float = 0.8,
+        batch_size: int = 16,
+        num_workers: int = 8,
+        prefetch_factor: int | None = None,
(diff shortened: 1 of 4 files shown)


viscy/data/tarrow.py
```diff
@@ -0,0 +1,176 @@
...
+        include_fov_names: list[str] = [],
+        train_samples_per_epoch: int = 100000,
+        val_samples_per_epoch: int = 10000,
+        resolution: int = 0,
+        z_slice: int = 0,
+        **kwargs,
+    ):
+        """Initialize TarrowDataModule.
+
+        Args:
```
ziw-liu, line 28: See #133

viscy/data/tarrow.py
```diff
@@ -0,0 +1,306 @@
...
+from typing import Callable, Sequence
+
+import numpy as np
+import torch.nn as nn
+from iohub.ngff import Position, open_ome_zarr
+from lightning.pytorch import LightningDataModule
+from tarrow.data.tarrow_dataset import TarrowDataset
+from torch.utils.data import ConcatDataset, DataLoader
+
+from viscy.utils.engine_state import set_fit_global_state
```
mattersoflight, line 11: `viscy.utils.engine_state` doesn't exist. Missing commit?

mattersoflight:
@edyoshikun I was able to draw a batch and construct the model! 
I now realize the model encodes temporal changes into a dense representation at each time point. The output tensor has 16 channels for each pixel in the input. In other words, the model "lifts" the image into a higher dimension instead of "compressing" it into an embedding.
image

For any meaningful comparison with dynaCLR, we will need a downstream embedding model:

image

PR #222, opened 2025-02-08
Knowledge distillation between channels

ziw-liu:
Using a fluorescence feature extractor's output as target for a label-free feature extractor, e.g. learn label-free infection classification from paired viral sensor images.

Also redefine some more MONAI transforms for use with CLI.


edyoshikun (approved):
this all LGTM! nice to have these implementations!

PR #225, opened 2025-02-13
Expose the option to not return negative samples in the triplet data module

### LINKED ISSUE #224: Triplet Dataset always return the negative sample
#204 added a `return_negative` parameter for `TripletDataset`. But this is not exposed in the `TripletDataModule` so there is no way to configure it from a user's perspective. In practice the negative sample is still processed, slowing down the training. This parameter also lacks documentation.

edyoshikun:
This PR fixes #224. 
- Added relevant doctrings
- Exposed variable in `TripletDataModule` so that it can configure the `TripletDataset` for `fit`.

### DIFF
--- a/viscy/data/triplet.py
+++ b/viscy/data/triplet.py
@@ -98,6 +98,8 @@ class TripletDataset(Dataset):
             by default "any"
             (sample negative from another track any time point
             and use the augmented anchor patch as positive)
+        return_negative : bool, optional
+            Whether to return the negative sample, by default True
         """
         self.positions = positions
         self.channel_names = channel_names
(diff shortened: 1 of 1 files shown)


viscy/data/triplet.py
```diff
@@ -330,6 +333,8 @@ def __init__(
             Future time interval to sample positive and anchor from,
             "any" means sampling negative from another track any time point
             and using the augmented anchor patch as positive), by default "any"
+        return_negative : bool, optional
+            Whether to return the negative sample, by default True
```
ziw-liu, line 337:
```suggestion
            Whether to return the negative sample during the fit stage
            (can be set to False when using a loss function like NT-Xent),
            by default True
```

ziw-liu (approved):
Thanks!

PR #228, opened 2025-03-04
Script to use ConvNeXt v1 Tiny as a feature extractor

ziw-liu:
Use ImageNet-pretrained encoder from timm as feature extractor.

For example, the PCA of the features (ALFI test split) shows separation between FOVs:

image


ziw-liu:
@Soorya19Pradeep Have you tried training a linear classifier for mitosis with these features?

Soorya19Pradeep:
I just added a section to the script with the linear classifier which will classify the cell state of interphase vs mitosis and the accuracy was computed to be 66.5%. 
Thanks for the comment @ziw-liu !

Soorya19Pradeep (approved):

PR #229, opened 2025-03-06
Check if the sliding window size is valid

ziw-liu:
Provide a better error message when the requested Z window size is larger than the image array.


talonchandler (approved):
LGTM!

PR #230, opened 2025-03-20
Fix dice score

ziw-liu:
Account for breaking change in https://github.com/Lightning-AI/torchmetrics/pull/3017.

Note that this PR uses a _different_ implement of Dice. The output is expected to differ. Treat this as a breaking change if you used Dice in prior work, and do not compare results between reports generated by different versions.


Soorya19Pradeep (approved):
I was able to import and use the dice metrics by updating my viscy environment to use this branch.

edyoshikun (approved):
LGTM! 

Thanks for catching this. I also was able to run the example.

PR #234, opened 2025-04-01
Testing MLFlow for CZI

edyoshikun:
This PR was used for testing the deployment of Cytoland models using MLFLow


edyoshikun:
this was unnecessary. If needed we can just use the MLFlow Logger https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.mlflow.html

PR #235, opened 2025-04-03
Add DOI badge

ziw-liu:


edyoshikun (approved):
nice! Thanks for this! we should add it to all the other repos.

Issue #237, opened 2025-04-04
Question about batch and error appearing


shannonhandley:
Hi, I am currently doing the Image translation (Virtual Staining) dlmbl exercise, however with my own dataset. My Zarr file is structured the same way as the demo.
 
When I run this line : batch = next(iter(train_dataloader))
This is the error I am getting: 

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\ AppData\Local\anaconda3\envs\viscy\lib\multiprocessing\spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "C:\Users\ AppData\Local\anaconda3\envs\viscy\lib\multiprocessing\spawn.py", line 126, in _main
    self = reduction.pickle.load(from_parent)
  File "C:\Users\ AppData\Local\anaconda3\envs\viscy\lib\site-packages\zarr\core.py", line 2425, in __setstate__
    self.__init__(**state)
TypeError: ImageArray.__init__() got an unexpected keyword argument 'store'

Would be great to have some advice on how to fix this. Not sure if it is an issue with my zarr data file, however 
Thanks

ziw-liu:
Hi @shannonhandley this sounds like an incompatibility between [iohub](https://github.com/czbiohub-sf/iohub) (the OME-Zarr library) and pickling on Windows. I have to admit that we never fully tested training on Windows. I have several suggestions before I can set up a Windows box to test:

1. Try setting all the `num_workers` arguments to `0`. This will turn off multi-processing and by extension pickling.
2. Try it on a Linux machine if you have access to one. Windows also supports a builtin Linux virtual machine (WSL) that can use the GPU, see [documentation](https://learn.microsoft.com/en-us/windows/ai/directml/gpu-cuda-in-wsl).

shannonhandley:
Thank you! Setting num_workers to 0 worked. However noting that I also got this error:

TypeError: default_collate: batch must contain tensors, numpy arrays, numbers, dicts or lists; found <class 'NoneType'>

For four of my image datasets it was returning None. Currently I am working around this by replacing it with zeros. Again not sure if this is an error from my side with the zarr file, but also in the process of trying it on a Linux machine as per your suggestion. 

Thanks again!

ziw-liu:
@shannonhandley the tutorial used a dataset that had pre-computed statistics. For your custom dataset did you run `viscy preprocess --data_path=your.hcs.ome.zarr`?

[...]

PR #238, opened 2025-04-10
Remove outdated docs

ziw-liu:
Remove outdated docs inherited from microdl.

Avoid confusions like https://github.com/mehta-lab/VisCy/issues/237#issuecomment-2791653817.


edyoshikun (approved):
I agree.

PR #241, opened 2025-04-17
Add evaluation script for infection classifier models

ziw-liu:
Add figure making script.


ziw-liu:
CI is broken on the target branch. Nothing we can do here...

PR #257, opened 2025-06-10
Debugging the TripletDataModule using the ConcatDataModule

edyoshikun:
This PR exposes some attributes existing in HCSDataloader that help speedup and reduce the the run-time, avoiding the timeout error.


ziw-liu:
@edyoshikun do you still need this?

PR #258, opened 2025-06-11
ONNX support for DynaCLR

edyoshikun:
This PR modifies the forward method to generat the features and projections.  

This change maintains compatibility with the existing contrastive learning pipeline while also making the encoder features accessible for other uses. The contrastive loss is still computed on the projections, but you now have access to the full feature representation when needed.


edyoshikun:
no need to modify the cli. It was more on the  forward method that it was returning the 32 dim tensor instead of the 768.

ziw-liu (approved):
Thanks @edyoshikun!

PR #260, opened 2025-06-16
Plot teacher accuracy

ziw-liu:
Update figure 4.


edyoshikun (approved):
Thanks @ziw-liu . Merging these.

PR #240, opened 2025-04-15
DynaCLR_V2 Submission

edyoshikun:
This PR contains the code of our revised dynaCLR pipeline.

As discussed, the goal is to merge this to `main` so we can have DynaCell branch can depend on some of these features. Additionally, we want these features to be ready prior to the preprint update. 

TODOS:
- [x] @edyoshikun will merge his DTW scripts
- [x] @edyoshikun will update the demo scripts to work with the current version of viscy
- [x] @edyoshikun update the README.md after #246 for the DynaCLR section
- [x] @edyoshikun review the demo README.md for typos.
- [x] @edyoshikun upload data to public and update the `setup.sh`
- [x] @edyoshikun @ziw-liu merge #250
- [x] @Soorya19Pradeep will merge and test the CellFeature #254 
- [x] @ziw-liu will review this PR. (You can give us review after Soorya implements the `CellFeatures`)

PR #262, opened 2025-06-17
Replace black with ruff

ziw-liu:
One less dependency.

Changed instructions for examples, but didn't apply to source files to avoid conflict with #240.


edyoshikun (approved):
LGTM! Thanks.

PR #267, opened 2025-07-02
Fix the PC Viewer to support multi-dataset PHATE embedding

edyoshikun:
This PR cleans up the PC viewer to support multi-dataset view of PHATE and PC.


Soorya19Pradeep:
Thank you @edyoshikun! 
I am getting the following error while using the script:

> Error processing batch for /B/2/000000, track [65]: unhashable type: 'list'
> Error processing batch for /B/2/000000, track [65]: unhashable type: 'list'
> Error processing batch for /B/2/000000, track [65]: unhashable type: 'list'
> Error processing batch for /B/2/000000, track [65]: unhashable type: 'list'
> Processing FOV /C/2/000001 with tracks all
> ['/C/2/000001', '/C/2/000001', '/C/2/000001']
> all
> Successfully cached 0 images
> Cached datasets: set()
> Cached dataset-FOV combinations: 0
> Number of unique dataset-FOV-track combinations: 0
> Saving image cache to /hpc/mydata/soorya.pradeep/scratch/cache/image_cache.pkl
> Successfully saved cache with 0 images

ziw-liu:
@Soorya19Pradeep was the freezing you observed happening on this branch?

Soorya19Pradeep:
> @Soorya19Pradeep seems like this is now working. Can we close and merge this PR to main?

Thanks @edyoshikun . I currently just have one issue with the widget which has a very easy workaround. The plotly widget stores information in my web browser, which leads to it breaking if I run the same code again. For instance, when I use the lasso tool to annotate clusters once in a specific PC space, then I close down the port. Then rerun the code and try lasso again, it would not do a great job and shows a combination of something I am trying to do and from the past annotations. It is not very reproducible. I read that this is a feature from Plotly, and I can open a new browser window to fix this. I was just unsure if there is a workaround for that through the code.

PR #268, opened 2025-07-09
Use roc auc instead of accuracy

ziw-liu:
Update metrics for knowledge distillation plots to be robust to class imbalance.


edyoshikun:
stale and addressed in #373

PR #278, opened 2025-07-27
Update the DLMBL Notebook

edyoshikun:
This updates the DLMBL to use viscy > 0.3.2 and upgrades to cellpose-sam


examples/virtual_staining/dlmbl_exercise/solution.py
```diff
@@ -1245,20 +1240,29 @@ def process_image(image):
 # - The segmentations will be stored in the `segmentation_store` zarr file
 # - Analyze the code while it runs.
 # %%
+# Create cellpose model once for reuse
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+cellpose_model = models.CellposeModel(gpu=True, device=device)
```
ziw-liu, line 1245: Does `gpu=True` work with CPU?

ziw-liu (approved):

PR #276, opened 2025-07-25
Batched dataloading for contrastive learning

ziw-liu:
## Changes

- Improve throughput of `TripletDataModule`
  - Use tensorstore for batched reading and decoding of patches
  - Move to a device-agnostic batched transform interface
- Implement transforms that run efficiently on the GPU
  - Batched normalization transforms for both pre-computed global statistics and on-the-fly percentiles scaling
  - Batched affine transform
  - Torch-based RNG for the Gaussian noise transform
- Batched version of `ConcatDataModule`

## Results

Training on a concatenation of 2 data modules from @Soorya19Pradeep’s script takes 6 min/epoch on a small node (16 zen 2 cores, 1 A40 GPU).

Example script:

```
/hpc/projects/organelle_phenotyping/models/SEC61_TOMM20_G3BP1_Sensor/time_interval/dynaclr_gfp_rfp_Ph/profiling/fit.py
```

image

PR #282, opened 2025-08-19
Batched transforms

ziw-liu:
Add more batched transforms.

Tested (now includes all DynaCLR transforms, along with #276):
- Random adjust contrast
- Random spatial crop
- Center spatial crop
- Random flip
- Random Gaussian noise
- Random scale intensity

See an illustration of these in the new example notebook.

Not tested (#284):
- Random 3D elastic
- Random histogram shift
- Random local pixel shuffle
- Random sharpen
- Random Z stack shift

Also updated the triplet dataset and datamodule for end-to-end batching. Now the dataset only handles I/O (in thread workers), and all the transforms are applied to the batched tensor on the GPU in the main thread. Since tensorstore manages its own thread pool, it is recommended to set `num_workers` to 1.


viscy/data/triplet.py
```diff
@@ -572,4 +566,46 @@ def predict_dataloader(self):
...
+        return self._augmentation_transform
+
+    def on_after_batch_transfer(self, batch, dataloader_idx: int):
+        """Apply transforms after transferring to device."""
+        if isinstance(batch, Tensor):
+            # example array
+            return batch
+        for key in ["anchor", "positive", "negative"]:
+            if key in batch:
+                norm_meta_key = f"{key}_norm_meta"
```
edyoshikun, line 599: here we are enforcing that all the dataset always run pre-processing. Should that be always true?

ziw-liu: For our current use case I think that's always true. We can change this if a need arises.

viscy/data/triplet.py
```diff
@@ -404,7 +403,13 @@ def __init__(
             Whether to pin memory in CPU for faster GPU transfer, by default False
         z_window_size : int, optional
             Size of the final Z window, by default None (inferred from z_range)
+        cache_pool_bytes : int, optional
+            Size of the per-process tensorstore cache pool in bytes, by default 0
         """
+        if num_workers > 1:
+            warnings.warn(
+                "Using more than 1 thread worker will likely degrade performance."
```
edyoshikun, line 411: We should add something more informative of why either in the docstrings or here. I prefer the docstrings if we are using batchedtransforms

edyoshikun: and we are not aiming for backwards compatibility then rename the whole argument

viscy/data/cell_classification.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         transform: Callable | None,
         initial_yx_patch_size: tuple[int, int],
         return_indices: bool = False,
+        label_column: str = "infection_state",
```
edyoshikun, line 26: I guess we are being flexible here, but this opens the door for non standarization.

viscy/data/triplet.py
```diff
@@ -143,14 +133,43 @@ def __init__(
         )
         self.valid_anchors = self._filter_anchors(self.tracks)
         self.return_negative = return_negative
+        self._setup_tensorstore_context(cache_pool_bytes)
+
+    def _setup_tensorstore_context(self, cache_pool_bytes: int):
+        """Configure tensorstore context with CPU limits based on SLURM environment."""
+        cpus_per_task = os.environ.get("SLURM_CPUS_PER_TASK")
```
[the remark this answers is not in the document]
ziw-liu: It only works when the process is spawned by `srun`, i.e. in the batch script.

viscy/data/triplet.py
ziw-liu: Added to docstring.

viscy/data/cell_classification.py
ziw-liu: Is developing the controlled vocabulary in scope for this PR?

edyoshikun: I made a new PR #300

edyoshikun (commented):
we dont have tests for some of the transforms (ie 3d elastic, histogram shift, local pixel shuffle, sharpen, and z_stack shift). 

Isn't the z_stack shift just an affine?

edyoshikun (commented):
I couldn't find a better way to optimize the funciotns and I really liked the tests.

ziw-liu:
> we dont have tests for some of the transforms (ie 3d elastic, histogram shift, local pixel shuffle, sharpen, and z_stack shift).
> 
> Isn't the z_stack shift just an affine?

These were contributions from #284.

edyoshikun (approved):

PR #285, opened 2025-08-25
3D ResNet for contrastive learning

ziw-liu:


edyoshikun:
@ziw-liu could you fix the merge conflicts? also there seems to be something broken with the diff

PR #300, opened 2025-09-16
Adding annotation typings

edyoshikun:
This PR takes a first stab at standardizing the labels and annotation columns for our datasets with the hope to merge annotations across datasets easily.

### DIFF
--- a/viscy/data/typing.py
+++ b/viscy/data/typing.py
@@ -97,11 +97,7 @@ AnnotationColumns = Literal[
 
 
 # NOTE: The following labels are not mutable. They are used to map the labels to the integer values.
-INFECTION_STATE_LABEL = {  
-    "uninfected": 0 ,
-    "infected": 1 ,
-    "unknown": -1  
-}
+INFECTION_STATE_LABEL = {"uninfected": 0, "infected": 1, "unknown": -1}
 
 CELL_CYCLE_PHASE_LABEL = {
     "interphase": 0,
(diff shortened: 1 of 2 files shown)


viscy/data/typing.py
```diff
@@ -88,3 +88,28 @@ class TripletSample(TypedDict):
...
+AnnotationColumns = Literal[
+    "infection_state", "cell_cycle_phase", "cell_remodeling_state"
+]
+
+
+# NOTE: The following labels are not mutable. They are used to map the labels to the integer values.
+class InfectionStateLabel(dict):
+    uninfected: 0
+    infected: 1
+    unknown: -1
```
ziw-liu, line 103: Class attributes of the dict type are not the same as key-values in dict. Use concrete objects instead:

```suggestion
INFECTION_STATE_LABEL = {
    "uninfected": 0
    "infected": 1
    "unknown": -1
    }
```

ziw-liu: Also since these are not types, they should be in a separate module.

viscy/data/typing.py
```diff
@@ -88,3 +88,29 @@ class TripletSample(TypedDict):
     positive: NotRequired[Tensor]
     negative: NotRequired[Tensor]
     index: NotRequired[TrackingIndex]
+
+
+# NOTE: these are the only columns that are allowed for the annotation dataframe.
+AnnotationColumns = Literal[
+    "infection_state", "cell_cycle_phase", "cell_remodeling_state"
```
ziw-liu, line 95: Is this the final vocabulary set?

srivarra (commented):
Looks good!

ziw-liu (approved):

PR #306, opened 2025-09-19
FID code from Dihan

edyoshikun:


.gitignore
```diff
@@ -47,3 +47,5 @@ slurm*.out
 
 #lightning_logs directory
 lightning_logs/
+
+applications/dynacell/test
```
ziw-liu, line 51: Is this intentional?

Issue #291, opened 2025-09-10
Lightning 2.5.5 breaks checkpoint reuse


ziw-liu:
We have workflows where checkpoints trained with one wrapper class is reused for another. For example, a `FullyConvolutionalMAE` wrapped in `FcmaeUNet` during training could be wrapped in `AugmentedPredictionVSUNet` for inference. https://github.com/Lightning-AI/pytorch-lightning/pull/21116 shipped in lightning 2.5.5 breaks this use case as it raises an exception if the hyperparameter does not match what's saved in the checkpoint.

PR #310, opened 2025-09-22
Pin lightning for 0.3

ziw-liu:
A stop gap for #291.


edyoshikun (approved):
LGTM. We should update the fcmae

PR #274, opened 2025-07-22
add percentiles to preprocessing

edyoshikun:

PR #317, opened 2025-10-02
add 5th and 95th percentile

ritvikvasan:
Adding percentile calculation to the preprocessing/normalization function

### DIFF
--- a/viscy/utils/mp_utils.py
+++ b/viscy/utils/mp_utils.py
@@ -260,11 +260,16 @@ def get_val_stats(sample_values):
                                         indices
     :return dict meta_row: Dict with intensity data for image
     """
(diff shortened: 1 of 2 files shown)


ziw-liu:
Related: #274

viscy/utils/mp_utils.py
```diff
@@ -260,11 +260,16 @@ def get_val_stats(sample_values):
                                         indices
     :return dict meta_row: Dict with intensity data for image
     """
+    p5 = float(np.nanpercentile(sample_values, 5))
+    p95 = float(np.nanpercentile(sample_values, 95))
```
ziw-liu, line 264: We can probably compute `q=[1, 5, 25, 50, 75, 95, 99]` in one percentile (sorting) call here and derive the median and iqr too.

PR #332, opened 2025-10-30
Log the similarity between anchor and negatives

edyoshikun:
This PR adds:
- Logs the cosine and Euclidean distance between the anchor and the negatives.
- Logs the margin between the cosine similarity between anchor-positive sample and the mean of anchor-negative sample. 
- Tests for the `_valid_anchors()` to ensure we are getting the right anchor+positive.


mattersoflight (approved):
LGTM on reading the files and can do "integration test" with other PRs later.

PR #334, opened 2025-11-03
Add precommit to dev dependencies

alxndrkalinin:
Currently `pre-commit` is not included in the development dependencies group, so the workflow described in [CONTRIBUTING.md](https://github.com/mehta-lab/VisCy/blob/main/CONTRIBUTING.md) does not work


edyoshikun (approved):
Thanks!! LGTM

PR #345, opened 2025-12-03
compute normalization per timepoint

ritvikvasan:
Update HCS preprocess script to save norm stats per timepoint in the zarr json


edyoshikun (approved):
I added some tests that we didn't have for these functions. I think we are good to go assuming they all pass. Thanks @ritvikvasan . Feel free to merge it if everything passes

PR #349, opened 2025-12-12
demo annotator for organelle box

edyoshikun:
This PR adds a simple annotator of cell states and returns ultrack csv files

### DIFF
--- a/examples/DynaCLR/annotation/napari_annotate.py
+++ b/examples/DynaCLR/annotation/napari_annotate.py
@@ -1,5 +1,7 @@
 # %%
 import logging
+import os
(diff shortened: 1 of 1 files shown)


edyoshikun:
@Soorya19Pradeep pointed out we need a way to version the annotations. I tested earlier today to use the `os.userlogin()` which spits out ids for each one of us and datetimes. To resume or edit existing annotations, we will still need to add some wrappers potentially from `ultracks napari plugin`.

edyoshikun:
@srivarra should we consider moving this to `napari-iohub` instead? there is a lot of code duplication.

PR #365, opened 2026-02-12
Modularizing the model's folder

edyoshikun:
- Making the modular `viscy-models` package. 
- restructure our model structure for the models. See README.md for details
- simple test for each architecture

Ignore the `.planning` folder. this is just so that we can sync them later to continue the refactor. We bring some of it to the CLAUDE.md but most likely delete most of it. once we are done with the staging branch

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,42 @@
+# CLAUDE.md
+
+Project-specific instructions for Claude Code sessions in this repository.
+
+## Repository Structure
+
+VisCy is a **uv workspace monorepo**. Sub-packages live under `packages/`:
+
+```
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
srivarra: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
Once we get to the final stage, we should figure how to independently run tests for a particular package if say changes are isolated to that package (assuming this makes sense at that point).

srivarra (approved):
this looks great

srivarra (approved):
The CLAUDE.md file looks good.

PR #366, opened 2026-02-12
modularizing data package

alxndrkalinin:
## Summary

Extract `viscy/data/` into an independent `viscy-data` package within the uv workspace monorepo (Milestone v1.1).

- **15 modules** migrated from `viscy.data.*` to `viscy_data.*` with clean break on imports
- **45 public exports** via PEP 562 lazy-loading `__init__.py` (optional deps load on demand)
- **4 optional dependency groups**: `triplet`, `livecell`, `mmap`, `all`
- **71+ tests** covering all DataModules, Datasets, and utilities
- **Tiered CI**: `test-data` (base, 3x3 OS/Python matrix) + `test-data-extras` (optional deps)

## Key decisions

- `from viscy_data import X` — clean break, no backward compatibility shims
- Lazy submodule loading via `__getattr__` — modules with optional deps (triplet, livecell, mmap_cache) only imported on access
- `_select.py` (private), `_typing.py` (private), `_utils.py` (private) — internal helpers not re-exported
- `ULTRACK_INDEX_COLUMNS` renamed from `INDEX_COLUMNS` for clarity
- `BatchedCenterSpatialCropd` removed from viscy-data (lives in viscy-transforms)
- Mutable default arguments replaced with `None` across all DataModules
- `NormalizeSampled` keys use `"source"` (not channel names) when applied via `HCSDataModule.normalizations`

## Bug fixes during migration

- `LiveCellTestDataset.__getitem__`: transform result was discarded (assigned but not returned)
- `ClassificationDataModule.setup()`: `raise (f"...")` raised a string instead of `ValueError`
- `CachedOmeZarrDataModule._filter_fit_fovs`: override silently ignored `exclude_fovs`
- `CTMCv1DataModule` and `LiveCellDataModule`: missing `prefetch_factor` attribute
- `CachedOmeZarrDataset`/`MmappedDataset`: used leaked loop variable for `self.channels`
- `LiveCellTestDataset`: `torch.stack` on empty list when no annotations
- `CombinedDataModule.setup()`: constituent DMs set up before stage validation
- Guard `None` dereferences in `_read_norm_meta`, `MaskTestDataset`, normalize metadata

## Phases

| Phase | Plans | What it built |
|-------|-------|---------------|
| 06: Package Scaffolding | 2 | pyproject.toml, type definitions, utility extraction, workspace registration |
| 07: Code Migration | 4 | Core modules (hcs, gpu_aug, select), specialized (triplet, classification), optional deps (livecell, mmap, ctmc), public API exports |
| 08: Test Migration | 2 | Conftest fixtures, 71 smoke + functional tests |
| 09: CI Integration | 1 | GitHub Actions test-data + test-data-extras jobs |

## Test plan

- [x] `uv run pytest packages/viscy-data/ -v` — all tests pass
- [x] `uvx ruff check packages/viscy-data/` — clean
- [x] `from viscy_data import HCSDataModule, TripletDataModule, ...` — all 45 exports work
- [x] Optional deps raise clear `ImportError` when missing
- [x] CI matrix: ubuntu/macos/windows x Python 3.11/3.12/3.13

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/viscy-data/tests/test_cell_division_triplet.py
+++ b/packages/viscy-data/tests/test_cell_division_triplet.py
@@ -1,0 +1,202 @@
+"""Smoke tests for CellDivisionTripletDataset and CellDivisionTripletDataModule."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import numpy as np
+from pytest import TempPathFactory, fixture, raises
+
+from viscy_data import CellDivisionTripletDataModule, CellDivisionTripletDataset
+
+
+@fixture(scope="function")
+def npy_data_dir(tmp_path_factory: TempPathFactory) -> Path:
+    """Create a temporary directory with several .npy cell-division track files.
+
+    Each file has shape (T, C, Y, X) with 2 channels (bf, h2b) like the real data.
+    """
+    data_dir = tmp_path_factory.mktemp("cell_div_npy")
+    rng = np.random.default_rng(42)
+    for i in range(6):
+        # Vary timepoints so filtering logic is exercised
+        t = rng.integers(3, 8)
+        arr = rng.random((t, 2, 32, 32)).astype(np.float32)
+        np.save(data_dir / f"track_{i}.npy", arr)
+    return data_dir
+
+
+# -- Dataset-level tests --
+
+
+def test_dataset_getitem_fit(npy_data_dir):
+    """__getitem__ returns anchor/positive/negative in fit mode."""
+    npy_files = sorted(npy_data_dir.glob("*.npy"))
+    ds = CellDivisionTripletDataset(
+        data_paths=npy_files,
+        channel_names=["bf", "h2b"],
+        fit=True,
+        return_negative=True,
+        time_interval="any",
+    )
+    assert len(ds) > 0
+    sample = ds[0]
+    assert "anchor" in sample
+    assert "positive" in sample
+    assert "negative" in sample
+    # Default 3D output: (C, 1, Y, X)
+    assert sample["anchor"].ndim == 4
+    assert sample["anchor"].shape[0] == 2  # two channels
+    assert sample["anchor"].shape[1] == 1  # depth dim
+
+
+def test_dataset_getitem_predict(npy_data_dir):
+    """In predict mode (fit=False) only anchor and index are returned."""
+    npy_files = sorted(npy_data_dir.glob("*.npy"))
+    ds = CellDivisionTripletDataset(
+        data_paths=npy_files,
+        channel_names=["bf"],
+        fit=False,
+        time_interval=1,
+    )
+    assert len(ds) > 0
+    sample = ds[0]
+    assert "anchor" in sample
+    assert "index" in sample
+    assert "positive" not in sample
+
+
+def test_dataset_output_2d(npy_data_dir):
+    """output_2d=True produces (C, Y, X) tensors."""
+    npy_files = sorted(npy_data_dir.glob("*.npy"))
+    ds = CellDivisionTripletDataset(
+        data_paths=npy_files,
+        channel_names=["bf"],
+        fit=True,
+        return_negative=True,
+        time_interval="any",
+        output_2d=True,
+    )
+    sample = ds[0]
+    assert sample["anchor"].ndim == 3  # (C, Y, X)
+
+
+def test_dataset_time_interval_filter(npy_data_dir):
+    """Non-'any' time_interval should filter valid anchors."""
+    npy_files = sorted(npy_data_dir.glob("*.npy"))
+    ds_any = CellDivisionTripletDataset(
+        data_paths=npy_files,
+        channel_names=["bf"],
+        fit=True,
+        time_interval="any",
+    )
+    ds_int = CellDivisionTripletDataset(
+        data_paths=npy_files,
+        channel_names=["bf"],
+        fit=True,
+        time_interval=2,
+    )
+    # With a concrete interval, fewer anchors should be valid
+    assert len(ds_int) < len(ds_any)
+
+
+def test_dataset_no_return_negative(npy_data_dir):
+    """return_negative=False omits the negative key."""
+    npy_files = sorted(npy_data_dir.glob("*.npy"))
+    ds = CellDivisionTripletDataset(
+        data_paths=npy_files,
+        channel_names=["bf"],
+        fit=True,
+        return_negative=False,
+        time_interval="any",
+    )
+    sample = ds[0]
+    assert "positive" in sample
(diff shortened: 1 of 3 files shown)


edyoshikun (commented):
there are some datasets that don't have a test.  (i.e ctmc, livevell, etc)

packages/viscy-data/README.md
```diff
@@ -0,0 +1,5 @@
+# viscy-data
+
+Data loading and Lightning DataModules for virtual staining microscopy.
```
edyoshikun: in the other one I put for `AI x Imaging`  tasks

packages/viscy-data/pyproject.toml
```diff
@@ -0,0 +1,65 @@
+[build-system]
+build-backend = "hatchling.build"
+requires = [ "hatchling", "uv-dynamic-versioning" ]
+
+[project]
+name = "viscy-data"
+description = "Data loading and Lightning DataModules for virtual staining microscopy"
```
edyoshikun: this also needs to be updated.

packages/viscy-data/src/viscy_data/cell_classification.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def _subset(
+        self,
+        plate: Plate,
+        annotation: "pd.DataFrame",
+        fov_names: list[str],
+        transform: Callable | None,
+        exclude_timepoints: list[int] = [],
+        return_indices: bool = False,
+    ) -> ClassificationDataset:
```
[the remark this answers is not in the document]
alxndrkalinin: Fixed in `71e94cf` — default changed to `None`, with an explicit guard in the method body.

packages/viscy-data/src/viscy_data/gpu_aug.py
```diff
@@ -0,0 +1,321 @@
...
+    def _filter_fit_fovs(self, plate: Plate) -> list[Position]:
+        """Filter FOVs from HCS plate for fitting."""
+        positions = []
+        for well_name, well in plate.wells():
+            if self._include_well_name(well_name):
+                for _, p in well.positions():
+                    positions.append(p)
+        if len(positions) < 2:
+            raise ValueError("At least 2 FOVs are required for training and validation.")
+        return positions
```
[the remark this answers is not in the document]
alxndrkalinin: Fixed in `4da09a0` — removed the local `_filter_fit_fovs` override so the `SelectWell` mixin's implementation (which filters both wells and FOVs) is used instead.

.github/workflows/test.yml
```diff
@@ -40,10 +40,61 @@ jobs:
         run: uv run --frozen pytest --cov=src/ --cov-report=term-missing
         working-directory: packages/${{ matrix.package }}
 
+  test-data:
```
srivarra: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
Ah, so you want to have separate testing jobs for each package. I think when we put the whole thing together we can have a matrix of test jobs for python version x os x package.

packages/viscy-data/src/viscy_data/_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        super().__init__(keys, cropper=cropper, allow_missing_keys=allow_missing_keys)
+
+
+def _ensure_channel_list(str_or_seq: str | Sequence[str]) -> list[str]:
+    """
+    Ensure channel argument is a list of strings.
+
+    :param Union[str, Sequence[str]] str_or_seq: channel name or list of channel names
+    :return list[str]: list of channel names
+    """
```
srivarra: [todo(blocking)](https://pullpo.io/cc?l=todo&d=blocking) 

Can we make these numpydocs?

packages/viscy-data/src/viscy_data/livecell.py
```diff
@@ -0,0 +1,302 @@
...
+            self._setup_fit()
+        elif stage == "test":
+            self._setup_test()
+
+    def _parse_image_names(self, annotations: Path) -> list[Path]:
+        """Parse image file names from COCO annotations."""
+        with open(annotations) as f:
+            images = [f["file_name"] for f in json.load(f)["images"]]
+        return sorted(images)
+
```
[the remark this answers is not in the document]
alxndrkalinin: Fixed in 82a420a — corrected the return type annotation from `list[Path]` to `list[str]`.

packages/viscy-data/src/viscy_data/triplet.py
```diff
@@ -0,0 +1,597 @@
...
+            [
+                track[(track["t"] + self.time_interval).isin(track["t"])]
+                for (_, track) in tracks.groupby("global_track_id")
+            ]
+        )
+
+    def _specific_cells(self, tracks: "pd.DataFrame") -> "pd.DataFrame":
+        """Filter tracks to specific cells by FOV name and track ID."""
+        specific_tracks = pd.DataFrame()
+        print(self.include_fov_names)
```
edyoshikun: _logger.debug(self.include_fov_names)
_logger.debug(self.include_track_ids)

alxndrkalinin:
> not sure what the `py.typed` file is for

Marker file per the [typing spec](https://typing.readthedocs.io/en/latest/source/libraries.html) to allow type checkers to use the package's inline annotations when it's downstream imported as an dependency. Also added an __init__.pyi type stub to support the new lazy-loading __init__.py.

> no strong opiions but the `viscy_data/select.py` is almost like just private functions and a dataclass. we can rename it to `_select.py`. right now based on the folder structure the rest are torch datasets

Renamed.

> there are some datasets that don't have a test. (i.e ctmc, livevell, etc)

I added some trying to follow similar patterns as the pre-existing tests (test_hcs.py, test_triplet.py).

>  `BatchedCenterSpatialCropd` belongs in `viscy-transforms`

Removed from `viscy-data` entirely. It already exists in `viscy-transforms`. Replaced the final crop in TripletDataModule with a shape validation check in `on_after_batch_transfer()`.

> CI jobs test-data and test-data-extras are identical

 Fixed. `test-data` now runs `uv sync --frozen --dev` (without `--all-extras`) so it validates the base package works without optional deps. `test-data-extras` keeps `--all-extras`.

>  `init.py` eagerly imports all modules including optional-dep ones

 Replaced eager imports with PEP 562 `__getattr__`-based lazy loading. Lightweight types (_typing, _select) stay eager; all DataModule/Dataset submodules are loaded on first access only. Added a subprocess-based smoke test that verifies no lazy submodule is in sys.modules after a bare import viscy_data. Added __init__.pyi stub for full IDE/type-checker support.

PR #395, opened 2026-03-28
Depend on waveorder>=3.0

ieivanov:


edyoshikun (approved):
@ieivanov note we will release a stable version of the current viscy and this will be replaced right after with the alpha from the staging monorepo branch

PR #450, opened 2026-06-03
feat(cytoland): non-square rotation TTA + reuse helpers

ieivanov:
## Summary

Make cytoland's prediction-time augmentation usable from external pipelines (e.g. biahub virtual staining) without re-implementing normalization or TTA.

- **Fix non-square rotation TTA.** `AugmentedPredictionVSUNet._predict_with_tta` cropped the prediction to the *original* spatial shape before applying the inverse transform. Shape-changing transforms (90°/270° rotations) swap Y and X, so for non-square FOVs the crop happened in the wrong frame and the per-rotation predictions could not be stacked/reduced. Now it crops to the **augmented** shape (`fwd_t(source).shape[2:]`) before `inv_t`. No square padding needed — works for any aspect ratio.
- **`rotation_tta_transforms(n=4)`** factory returning the standard 90° rotation `(forward, inverse)` transform lists.
- **`AugmentedPredictionVSUNet.with_rotation_tta(model, reduction="median")`** convenience constructor that wires the factory in, so callers don't hand-build transforms. (The library now uses its own helper.)
- **`viscy_data.read_norm_meta(fov)`** — public wrapper over `_read_norm_meta` for reading precomputed normalization stats from a FOV's `zattrs`.

## Test plan

- [x] New regression tests: `test_predict_sliding_windows_rotation_tta_nonsquare` (64×64, 64×48, 48×64) and `test_rotation_tta_transforms`.
- [x] `pytest test_engine.py -k "rotation_tta or nonsquare or sliding_windows"` → 8 passed.

### DIFF
--- a/applications/cytoland/src/cytoland/engine.py
+++ b/applications/cytoland/src/cytoland/engine.py
@@ -94,6 +94,8 @@ def rotation_tta_transforms(
     tuple[list[Callable], list[Callable]]
         The forward and inverse transform lists.
     """
+    if n < 1:
+        raise ValueError(f"n must be >= 1, got {n}")
     forward = [partial(torch.rot90, k=k, dims=(-2, -1)) for k in range(n)]
     inverse = [partial(torch.rot90, k=-k, dims=(-2, -1)) for k in range(n)]
     return forward, inverse
--- a/applications/cytoland/src/cytoland/engine.py
+++ b/applications/cytoland/src/cytoland/engine.py
@@ -3,6 +3,7 @@
 import inspect
 import logging
 import os
+from functools import partial
 from typing import Callable, Literal, Sequence
 
 import numpy as np
@@ -68,6 +69,34 @@ def _center_crop_to_shape(tensor: Tensor, spatial_shape: tuple[int, ...]) -> Ten
         start = (current - size) // 2
         slices[dim] = slice(start, start + size)
     return tensor[tuple(slices)]
+
+
+def rotation_tta_transforms(
+    n: int = 4,
+) -> tuple[list[Callable[[Tensor], Tensor]], list[Callable[[Tensor], Tensor]]]:
+    """Build forward/inverse 90-degree rotation transforms for test-time augmentation.
+
+    Returns ``(forward_transforms, inverse_transforms)`` suitable for
+    :class:`AugmentedPredictionVSUNet`. Each forward transform rotates the YX
+    plane by ``k * 90`` degrees (``k = 0 .. n-1``) and the matching inverse
+    rotates back. Combined with ``reduction="median"`` this reproduces the
+    rotation TTA used by :meth:`VSUNet.perform_test_time_augmentations`, and
+    (unlike passing rotations through other code paths) works for non-square
+    fields of view.
+
+    Parameters
+    ----------
+    n : int, optional
+        Number of 90-degree rotations, by default 4 (0, 90, 180, 270 degrees).
+
+    Returns
+    -------
+    tuple[list[Callable], list[Callable]]
+        The forward and inverse transform lists.
+    """
+    forward = [partial(torch.rot90, k=k, dims=(-2, -1)) for k in range(n)]
+    inverse = [partial(torch.rot90, k=-k, dims=(-2, -1)) for k in range(n)]
+    return forward, inverse
 
 
 class MaskedMSELoss(nn.Module):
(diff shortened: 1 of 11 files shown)


ieivanov:
Refactored `biahub virtual-stain` implementation: https://github.com/czbiohub-sf/biahub/pull/267

packages/viscy-data/src/viscy_data/_utils.py
```diff
@@ -136,7 +136,7 @@ def _collate_samples(batch: Sequence[Sample]) -> Sample:
     return collated
 
 
-def _read_norm_meta(fov: Position) -> NormMeta | None:
+def read_norm_meta(fov: Position) -> NormMeta | None:
```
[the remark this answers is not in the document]
ieivanov: Good point, I like the idea of `_read_norm_meta = read_norm_meta` to avoid more code changes than necessary

ieivanov:
@edyoshikun @alxndrkalinin it would be great if we can squeeze #450 and #451 in the upcoming VisCy release if you agree with the changes I've suggested. If you think you need more time for review, I'm happy if we merge these later also.

packages/viscy-data/src/viscy_data/_utils.py
```diff
@@ -165,6 +165,10 @@ def _read_norm_meta(fov: Position) -> NormMeta | None:
     return norm_meta
 
 
+# Backwards-compatible private alias: existing callers import ``_read_norm_meta``.
+_read_norm_meta = read_norm_meta
```
edyoshikun: In the future we should consider if this is the right approach. I'm ok for now to keep the alias.

edyoshikun: I would actually say to do the opposite since this is already a _utils.py file.

edyoshikun: so read_norm_meta = _read_norm_meta

applications/cytoland/src/cytoland/engine.py
```diff
@@ -70,6 +71,36 @@ def _center_crop_to_shape(tensor: Tensor, spatial_shape: tuple[int, ...]) -> Ten
...
+def rotation_tta_transforms(
+    n: int = 4,
+) -> tuple[list[Callable[[Tensor], Tensor]], list[Callable[[Tensor], Tensor]]]:
+    """Build forward/inverse 90-degree rotation transforms for test-time augmentation.
+
+    Returns ``(forward_transforms, inverse_transforms)`` suitable for
+    :class:`AugmentedPredictionVSUNet`. Each forward transform rotates the YX
+    plane by ``k * 90`` degrees (``k = 0 .. n-1``) and the matching inverse
+    rotates back. Combined with ``reduction="median"`` this reproduces the
+    rotation TTA used by :meth:`VSUNet.perform_test_time_augmentations`, and
```
edyoshikun, line 83: do we want just rotations or do you want to add flips in x and y?

edyoshikun: they are really fast operations but you add 4 more transforms. I'm guessing you dont need them right now.

edyoshikun (commented):
I think there is only one switch for the `read_norm_meta` . other than that this is good.

applications/cytoland/src/cytoland/engine.py
```diff
@@ -659,9 +724,15 @@ def _predict_with_tta(self, source: Tensor) -> Tensor:
         preds = []
         for fwd_t, inv_t in zip(self._forward_transforms, self._inverse_transforms):
             aug_source = fwd_t(source)
+            # Crop back to the augmented (post-forward-transform) spatial shape,
+            # not the original one: a shape-changing transform such as a 90/270
+            # degree rotation swaps Y and X, so the prediction lives in the
+            # augmented frame until ``inv_t`` undoes the transform. Cropping to
+            # ``source.shape[2:]`` here would be wrong for non-square inputs.
+            aug_shape = aug_source.shape[2:]
```
edyoshikun, line 732: aug_source.shape[-2:] is probly better since the last two dimensions are always the xy I believe

applications/cytoland/src/cytoland/engine.py
ieivanov: Claude said the current codebase (`viscy predict`) does only rotations, you here I implemented the same. Let's bundle flips in a separate PR? We could package as `with_rotation_tta`, `with_reflection_tta`, and `with_tta` - which allows you to pick between rotation, reflection, or both.

applications/cytoland/src/cytoland/engine.py
ieivanov: Here `shape[2:]` isn't undoing the rotation (the rotation only touches XY, as you say); it's undoing `_predict_pad` / `_make_divisible_pad`, which pads **Z too** when `model.downsamples_z` is true:

```python
return DivisiblePad((0, down_factor, down_factor, down_factor))  # pads Z, Y, X
```

So for a z-downsampling 3D model, `aug_source.shape[-2:]` would crop back Y and X but leave Z still padded. `shape[2:]` reverses all the padded spatial dims (Z, Y, X), and it stays consistent with `_pad_forward_crop`, which uses `source.shape[2:]` for the same reason. The non-square fix is just that we crop to the *augmented* shape rather than the original.

For 4D (2D model) inputs the two are identical anyway, so `shape[2:]` is correct in both cases. I'll leave it as `shape[2:]`.

packages/viscy-data/src/viscy_data/_utils.py
ieivanov: implemented in [2e9bf7b](https://github.com/mehta-lab/VisCy/pull/450/commits/2e9bf7b98d28b482bb2db3d58be18feca399b3c9)

applications/cytoland/src/cytoland/engine.py
ieivanov: Opened #462 to track this feature

ieivanov:
@edyoshikun please take another look, I think I've addressed all your comments here

applications/cytoland/src/cytoland/engine.py
edyoshikun: I mean the ideal case in stead of bloating the code is to just pass the composable list of transforms. The gain from TTA is minimal so we can just do the rotations for now.

edyoshikun (approved):
Thanks for the changes. LGTM. I think this unblocks you now? @ieivanov

PR #451, opened 2026-06-03
refactor(cytoland): unify VSUNet rotation TTA on rotation_tta_transforms

ieivanov:
Stacked on #450.

## Summary

`VSUNet.perform_test_time_augmentations` (the `viscy predict` TTA path) had its **own** rotation implementation — `_rotate_volume` (square-pad + monai `Rotate90`) + `_crop_to_original` — separate from `AugmentedPredictionVSUNet`'s transform-list path. This unifies both on the `rotation_tta_transforms()` helper added in #450, so rotation TTA has a single source of truth.

- `perform_test_time_augmentations` now builds its rotations from `rotation_tta_transforms()` and reuses `_pad_forward_crop`, which crops to the **augmented** shape before the inverse rotation — so non-square FOVs work with bare `torch.rot90`, no square padding.
- Removes `_rotate_volume`, `_crop_to_original`, the `_original_shape_yx` attribute, and the now-unused `Rotate90` import.
- `mean`/`median`/`product` reductions preserved.

## Behavior change (intended)

For **non-square** FOVs, `viscy predict` TTA output changes: the old path zero-padded to a square before rotating (introducing border regions that were cropped back out); the new path rotates without padding. **Square FOVs are unchanged.** This is a correctness improvement for non-square inputs.

## Test plan

- [x] New `test_vsunet_perform_test_time_augmentations_nonsquare` (64×64, 64×48, 48×64 × mean/median).
- [x] `pytest test_engine.py -k "tta or rotation or augment or sliding_windows"` → 15 passed.

### DIFF
--- a/applications/cytoland/src/cytoland/engine.py
+++ b/applications/cytoland/src/cytoland/engine.py
@@ -11,7 +11,7 @@ import torch
 import torch.nn.functional as F
 from imageio import imwrite
 from lightning.pytorch import LightningModule
-from monai.transforms import DivisiblePad, Rotate90
+from monai.transforms import DivisiblePad
 from torch import Tensor, nn
 from torchmetrics.functional import (
     accuracy,
@@ -226,7 +226,6 @@ class VSUNet(LightningModule):
         self.test_time_augmentations = test_time_augmentations
         self.tta_type = tta_type
         self.freeze_encoder = freeze_encoder
-        self._original_shape_yx = None
         if ckpt_path is not None:
             self.load_state_dict(torch.load(ckpt_path, weights_only=True, map_location="cpu")["state_dict"])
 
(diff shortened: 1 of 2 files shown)


ieivanov:
Ed, Alex, this is optional, but I think nice to have. It unifies how `AugmentedPredictionVSUNet` and `VSUNet` do TTA transforms. As a result, now `viscy predict` gives better results (at least according to Claude) for non-square inputs. Let me know what you think, happy to refactor or scrap it altogether if you think it's not needed.

ieivanov:
### ⚠️ Known failing test: `product` TTA produces NaN

`test_vsunet_perform_test_time_augmentations_nonsquare[product-*]` fails for all 3 FOV shapes (`yx0/yx1/yx2`). `mean` and `median` pass.

**Cause:** `tta_type="product"` aggregates predictions in log space (`torch.log(p + 1e-9)`). The test uses an untrained model with random weights, whose outputs are frequently negative → `log()` of a negative is `NaN`, so the `torch.isfinite(output).all()` assertion fails.

**Not introduced by the main merge** (`a31eaa45`) — confirmed identical failures on the pre-merge branch tip `cf1a4a71` ("add 'product' to tta tests"), where the `product` parametrization was first added.

**Suggested fix (separate from this PR):** either
- clamp/relu predictions to be non-negative before the log in `perform_test_time_augmentations`, or
- give the `product` case a fixture with positive outputs (the product method only makes sense for non-negative predictions anyway).

edyoshikun (approved):

PR #472, opened 2026-06-17
fix(viscy-utils): fall back to TensorBoard logger when wandb is absent

### LINKED ISSUE #471: viscy CLI crashes on base install: default trainer.logger requires optional wandb
## Summary

`viscy <subcommand>` crashes on a **base install** (without the optional `eval` extra) because `VisCyCLI` hard-codes the default `trainer.logger` to a `WandbLogger`, but `wandb` is only an optional dependency. Any subcommand that builds a `Trainer` — `predict`, `preprocess`, `export`, `precompute`, `convert_to_anndata` — fails during class instantiation with a "Requirement 'wandb>=0.12.10' not met" error.

## Reproduction

In an environment where `viscy-utils` is installed **without** the `eval` extra (so `wandb` is not present):

```bash
viscy predict -c predict_config.yml --data.init_args.data_path input.zarr/A/1/0
```

## Error

```
ValueError: Does not validate against any of the Union subtypes
Subtypes: [<class 'NoneType'>, <class 'lightning.pytorch.loggers.logger.Logger'>, collections.abc.Iterable[lightning.pytorch.loggers.logger.Logger], <class 'bool'>]
Errors:
  - Expected a <class 'NoneType'>
  - Problem with given class_path 'lightning.pytorch.loggers.WandbLogger':
      Requirement 'wandb>=0.12.10' not met. HINT: Try running `pip install -U 'wandb>=0.12.10'`
  - Expected a <class 'collections.abc.Iterable'>
  - Expected a <class 'bool'>
Given value type: <class 'jsonargparse._namespace.Namespace'>
Given value: Namespace(class_path='lightning.pytorch.loggers.WandbLogger', init_args=Namespace(...))
```

The failure happens in `LightningCLI.instantiate_classes()` → `instantiate_trainer()` while resolving the `trainer.logger` default.

## Root cause

`packages/viscy-utils/src/viscy_utils/cli.py`:

```python
def add_arguments_to_parser(self, parser) -> None:
    """Set default logger."""
    parser.set_defaults(
        {
            "trainer.logger": lazy_instance(WandbLogger),
        }
    )
```

This sets a `WandbLogger` as the default logger for **every** subcommand, unconditionally. But `wandb` is not a core dependency — in `packages/viscy-utils/pyproject.toml` it appears only under `optional-dependencies.eval` (and the `test` dependency group), not in `dependencies`. So a base install advertises a default that it cannot instantiate, and the CLI is unusable for inference/preprocessing without either installing the `eval` extra or manually overriding the logger.

Subcommands like `predict` and `preprocess` don't even use the logger, so requiring `wandb` for them is pure overhead on top of being a hard failure.

## Workarounds (for users hitting this today)

- Pass `--trainer.logger false` (or `--trainer.logger null`) on the command line, or
- Install the extra: `pip install 'viscy-utils[eval]'`.

## Suggested fix

Only default to `WandbLogger` when `wandb` is actually importable; otherwise leave Lightning's built-in default (TensorBoard — `tensorboard` is already a core dependency) in place. This keeps the W&B convenience for users who have the `eval` extra without breaking base installs.

```python
import importlib.util
...
    def add_arguments_to_parser(self, parser) -> None:
        """Default to the W&B logger only when wandb is installed.

        wandb lives in the optional `eval` extra, so a base install must not
        hard-require it. When it is absent, leave Lightning's default logger
        (TensorBoard, a core dependency) in place so every subcommand still runs.
        """
        if importlib.util.find_spec("wandb") is not None:
            parser.set_defaults({"trainer.logger": lazy_instance(WandbLogger)})
```

`before_instantiate_classes` → `_configure_wandb_logger` is already a no-op when the resolved logger isn't a `WandbLogger` (it checks `class_path`), so it stays safe with this change.

Alternative approaches, if a W&B default is desired regardless:
- Move `wandb` from the `eval` extra into core `dependencies` (heavier — pulls in gitpython/sentry-sdk/etc. for every install).
- Default to an always-available logger (e.g. `CSVLogger`/`TensorBoardLogger`) and let users opt into W&B via config.

## Environment

- viscy-utils `0.0.0.post211.dev0+7fb3bc1` (branch `feat/vs-nonsquare-tta-and-helpers`)
- lightning `2.6.5`
- Reproduced via `biahub virtual-stain`'s preprocessing step (`viscy preprocess`), which builds the same CLI/Trainer.

ieivanov:
## Summary

`VisCyCLI.add_arguments_to_parser` unconditionally set the default `trainer.logger` to a `WandbLogger`. But `wandb` is only an **optional** dependency (the `eval` extra / `test` group), not a core one. So on a base install, every trainer-building subcommand — `predict`, `preprocess`, `export`, `precompute`, `convert_to_anndata` — crashed during class instantiation with:

```
Problem with given class_path 'lightning.pytorch.loggers.WandbLogger':
    Requirement 'wandb>=0.12.10' not met.
```

## Fix

Pick the default logger at runtime:
- `WandbLogger` when `wandb` is importable (unchanged behavior for users with the `eval` extra),
- otherwise `TensorBoardLogger` (`tensorboard` is already a core dependency), so the CLI works on a base install.

`before_instantiate_classes` → `_configure_wandb_logger` already no-ops when the resolved logger isn't a `WandbLogger` (it checks `class_path`), so the TensorBoard fallback is safe.

## Tests

Added unit tests for the new `_default_logger` helper (monkeypatching `importlib.util.find_spec`) covering both branches. Full `test_cli.py` passes; `ruff check`/`ruff format` clean.

Closes #471

### DIFF
--- a/packages/viscy-utils/src/viscy_utils/cli.py
+++ b/packages/viscy-utils/src/viscy_utils/cli.py
@@ -1,7 +1,6 @@
 """VisCy Lightning CLI with custom defaults."""
 
 import atexit
-import importlib.util
 import logging
 import os
 import re
@@ -16,7 +15,7 @@ import yaml
 from jsonargparse import Namespace, lazy_instance
 from lightning.pytorch import LightningDataModule, LightningModule
 from lightning.pytorch.cli import LightningCLI
-from lightning.pytorch.loggers import TensorBoardLogger, WandbLogger
+from lightning.pytorch.loggers import WandbLogger
 
 from viscy_utils.compose import load_composed_config
 from viscy_utils.trainer import VisCyTrainer
@@ -24,20 +23,6 @@ from viscy_utils.trainer import VisCyTrainer
 _WANDB_LOGGER_CLASS_PATH = "lightning.pytorch.loggers.WandbLogger"
 _WANDB_RUN_NAME_PREFIX = re.compile(r"^\d{8}-\d{6}_")
 _WANDB_RUN_TIMESTAMP_FORMAT = r"%Y%m%d-%H%M%S"
-
-
-def _default_logger():
-    """Return the default ``trainer.logger``: W&B when installed, else TensorBoard.
-
-    ``wandb`` is an optional dependency (the ``eval`` extra), so a base install
-    must not default to a logger it cannot instantiate -- otherwise every
-    trainer-building subcommand (``predict``, ``preprocess``, ``export``, ...)
-    fails at class instantiation. ``tensorboard`` is a core dependency, so it is
-    a safe fallback that keeps the CLI usable without the extra.
-    """
-    if importlib.util.find_spec("wandb") is not None:
-        return lazy_instance(WandbLogger)
-    return lazy_instance(TensorBoardLogger, save_dir=".")
 
 
 def _prefix_wandb_run_name(base_name: str, run_timestamp: str) -> str:
@@ -99,10 +84,10 @@ class VisCyCLI(LightningCLI):
         return subcommands
 
     def add_arguments_to_parser(self, parser) -> None:
-        """Set default logger (W&B when installed, else TensorBoard)."""
+        """Set default logger."""
         parser.set_defaults(
             {
-                "trainer.logger": _default_logger(),
+                "trainer.logger": lazy_instance(WandbLogger),
             }
         )
 
--- a/packages/viscy-utils/src/viscy_utils/cli.py
+++ b/packages/viscy-utils/src/viscy_utils/cli.py
@@ -1,6 +1,7 @@
 """VisCy Lightning CLI with custom defaults."""
 
 import atexit
+import importlib.util
 import logging
 import os
 import re
@@ -15,7 +16,7 @@ import yaml
 from jsonargparse import Namespace, lazy_instance
 from lightning.pytorch import LightningDataModule, LightningModule
 from lightning.pytorch.cli import LightningCLI
-from lightning.pytorch.loggers import WandbLogger
+from lightning.pytorch.loggers import TensorBoardLogger, WandbLogger
 
 from viscy_utils.compose import load_composed_config
 from viscy_utils.trainer import VisCyTrainer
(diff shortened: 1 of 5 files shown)


srivarra (approved):

packages/viscy-utils/src/viscy_utils/cli.py
```diff
@@ -25,6 +26,20 @@
 _WANDB_RUN_TIMESTAMP_FORMAT = r"%Y%m%d-%H%M%S"
 
 
+def _default_logger():
```
edyoshikun: I think this fix is a bit dangerous because it won't fail. Intestead it will try to find the second option. I am going to fix this by just making wandb a dependency

edyoshikun (changes requested):
I added the changes. review and test @ieivanov if this works for your issue.

ieivanov:
Yup, that works too, thank you!

ieivanov:
## Why the macOS DDP tests are now skipped

The macOS (and fail-fast-canceled Windows) **Test Data** jobs were failing on 3 parametrizations of `test_combined_ddp.py::test_batched_concat_real_ddp_iter_does_not_hang` with `SIGSEGV`/`SIGABRT` in the `torch.multiprocessing` fork child.

The real cause (reproduced deterministically on macOS):

```
objc[...]: +[MPSGraphObject initialize] may have been in progress in another thread
when fork() was called. We cannot safely call it or ignore it in the fork() child
process. Crashing instead.
```

This is the macOS Objective-C / Metal (MPS) fork-safety abort. The test intentionally uses `start_method="fork"` (spawn can't re-resolve modules under pytest's `--import-mode=importlib`), and forks a **multi-threaded** parent. Once torch initializes the MPS runtime, `fork()` of that process is fatal on Darwin — it's a hard crash, not a flake.

This isn't caused by this PR's logic. The PR only moves `wandb` into core deps and regenerates `uv.lock` (torch stays at 2.12.0); `main` passed these tests by timing luck, and the dependency reshuffle tipped MPS init across the fork boundary so it now crashes deterministically.

**Fix:** the test already skips on Windows (`"fork" not in mp.get_all_start_methods()`); I mirrored that with a `sys.platform == "darwin"` skip. These real-fork DDP integration tests now run only on the **Linux CI runner**, which passes them and is where actual DDP training happens.

edyoshikun (approved):
I added the changes. review and test @ieivanov if this works for your issue.

PR #480, opened 2026-07-14
feat(dynacell/tools): canonical artifact-path migration + config codemod

alxndrkalinin:
Relocates the dynacell benchmark artifacts (trained checkpoints, prediction zarrs, eval outputs) onto the `dynacell.evaluation.paths` grammar and repoints the config leaves to match — the barrier before the A549 raw-flip re-predict / re-eval.

**Stacked on #479** (`dynacell-paths-module`), which provides `paths.py`. Review/merge #479 first.

> **Status: the storage migration has already been applied** (with data-owner + author sign-off): 663 moves (67 checkpoints / 269 predictions / 327 eval leaves), journaled and reversible. This PR is the tooling plus the git-tracked config codemod output that matches the on-disk result; nothing here re-runs a migration on merge.

### Canonical grammar (relative)
- checkpoint: `<train_set>/<organelle>/<model>/checkpoints`
- prediction: `<organelle>/<model>/<train_set>/<test>[__<cond>]/prediction.zarr`
- eval: `<organelle>/<model>/<train_set>/<test>[__<cond>]/[<component>/][instance_ap/]`

Prediction and eval share one leaf dir (metrics + `embeddings/` + `instance_ap/` alongside `prediction.zarr`). Checkpoints consolidate the former two-model-root split into a single root.

### What's here
- **`build_migration_manifest.py`** — enumerates checkpoints/predictions/evals, computes legacy→canonical moves, dedups hardlinked dual-homes, curates the iPSC experiment set to the paper's final recipes, writes a CSV manifest + report. Exits non-zero on gaps/collisions.
- **`run_migration.py`** — journaled apply (`--apply`), dry-run default, idempotent resume, `--rollback`, pre-flight aborts (nested/cross-device/pre-existing dest) before any mutation.
- **`restandardize_config_paths.py`** — predict/train leaf codemod; rewrite-iff-migrated, so configs and the on-disk migration agree by construction.
- **Applied config codemod** — 318 leaves / 930 fields repointed (228 `ckpt_path`, 252 `output_store`, 318 `run_root`, 66 `dirpath`, 66 `save_dir`); ER/mito A549 checkpoints → raw `a549` train_set, their legacy-deconv predictions → `a549__deconv`; training inputs / external-init / `REPLACE_ME` preserved.
- **Checkpoint filenames now embed val loss** — restored `epoch={epoch}-step={step}-loss={<monitor>:.3f}` + `auto_insert_metric_name: false` on 79 training leaves (68 `loss/validate`, 11 `loss/validate_ema`; no-monitor celldiff/smoke skipped), so `--ckpt best` targets are identifiable from the filename.

### Two bugs found + fixed during the apply
- **Eval-into-shared-prediction-leaf merge** (`c553063e`): the prediction move creates the shared leaf and drops `prediction.zarr` in first, so the eval whole-dir rename hit `FileExistsError` (dry-run never created the dir, so it was invisible; pre-flight only checked dest-under-src). Now the eval's children are merged into the leaf, journaled per-child so `--rollback` is exact and never touches `prediction.zarr`.
- **Codemod map from the frozen manifest** (`552b9963`): `build_ckpt_map` re-scanned the live models tree, but post-migration the legacy dirs are gone → empty map → silent no-op rewrites. Added `load_ckpt_map_from_manifest(--manifest)`.

### Tests
`test_migration_tools.py` (9, incl. merge + rollback) and `test_restandardize_config_paths.py` (8, incl. deconv split, cross-root, manifest map) pass. The filename template's `/`-containing metric parses cleanly through the fit CLI.

### Deferred (not in this PR)
Eval-leaf codemod (`io.pred_path` / `save_dir`) + the grouped/instance_ap generator-walker fix + retiring the zero-importer `save_paths.py` shim — coupled to the post-migration on-disk layout.

### DIFF
--- a/applications/dynacell/configs/benchmarks/virtual_staining/_dual_nucl_memb/vscyto3d_cytolandft/a549_mantis/train.yml
+++ b/applications/dynacell/configs/benchmarks/virtual_staining/_dual_nucl_memb/vscyto3d_cytolandft/a549_mantis/train.yml
@@ -48,6 +48,8 @@ trainer:
     - class_path: lightning.pytorch.callbacks.ModelCheckpoint
       init_args:
         monitor: loss/validate
+        filename: "epoch={epoch}-step={step}-loss={loss/validate:.3f}"
+        auto_insert_metric_name: false
         every_n_epochs: 1
         save_top_k: 5
         save_last: true
--- a/applications/dynacell/configs/benchmarks/virtual_staining/_dual_nucl_memb/vscyto3d_cytolandft/ipsc_confocal/train.yml
+++ b/applications/dynacell/configs/benchmarks/virtual_staining/_dual_nucl_memb/vscyto3d_cytolandft/ipsc_confocal/train.yml
@@ -48,6 +48,8 @@ trainer:
     - class_path: lightning.pytorch.callbacks.ModelCheckpoint
       init_args:
         monitor: loss/validate
+        filename: "epoch={epoch}-step={step}-loss={loss/validate:.3f}"
+        auto_insert_metric_name: false
         every_n_epochs: 1
         save_top_k: 5
         save_last: true
--- a/applications/dynacell/configs/benchmarks/virtual_staining/_dual_nucl_memb/vscyto3d_infectionft_dynacellft/a549_mantis/train.yml
+++ b/applications/dynacell/configs/benchmarks/virtual_staining/_dual_nucl_memb/vscyto3d_infectionft_dynacellft/a549_mantis/train.yml
@@ -48,6 +48,8 @@ trainer:
     - class_path: lightning.pytorch.callbacks.ModelCheckpoint
       init_args:
         monitor: loss/validate
+        filename: "epoch={epoch}-step={step}-loss={loss/validate:.3f}"
+        auto_insert_metric_name: false
         every_n_epochs: 1
         save_top_k: 5
         save_last: true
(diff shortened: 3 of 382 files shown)


alxndrkalinin:
### Code review

Found 1 issue:

1. `_rewrite_ckpt_value` locates the model dir by splitting on the literal `"/checkpoints"` substring, but legacy checkpoints stored as a bare best-epoch hardlink alias at the model run-dir root have no `/checkpoints/` segment. `value.find("/checkpoints")` then returns `-1`, so `model_dir` becomes the entire value **including the filename** — which never matches a directory key in `ckpt_map`. The lookup returns `None` and the codemod silently preserves the stale pre-migration path (contrary to the docstring's "rewrite-iff-migrated" contract). 12 committed predict leaves keep a `ckpt_path` pointing at a now-deleted directory and will fail with checkpoint-not-found when a predict job launches: `er/fcmae_vscyto3d_pretrained/ipsc_confocal/`, `membrane/fcmae_vscyto3d_scratch/ipsc_confocal/`, and `nucleus/fcmae_vscyto3d_scratch/ipsc_confocal/` (4 `predict__*.yml` each). Their sibling `train.yml` files were rewritten correctly, and the tests never exercise a `ckpt_path` without a `/checkpoints/` segment. (bug due to the `idx == -1` fallback: `model_dir = value[:idx] if idx != -1 else value`)

https://github.com/mehta-lab/VisCy/blob/9193d2309e49096788e58f1fb36b81334517e527/applications/dynacell/tools/restandardize_config_paths.py#L120-L126

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

alxndrkalinin:
Fixed in ff60edb2 (tool) + fa5d5fbd (applied output).

`_rewrite_ckpt_value` now splits on the final path separator when there is no `/checkpoints` tail, with an exact bare-model-dir map-key short-circuit first (so `save_dir` / train `run_root` values aren't truncated by `rpartition`). The 12 iPSC-fcmae predict leaves are repointed to their canonical dirs (each target verified present on disk), plus 2 celldiff smoke `run_root`s the same bug had suppressed. Regression test added (`test_rewrite_ckpt_run_root_alias_no_checkpoints_subdir`); 17/17 codemod+migration tests pass.

